PR #5, opened 2020-07-02
Setup lint config and .gitattributes

antoncoding:


antoncoding:
Rules for `ts` and `js` files: 
`.eslintrc.json` && `prettierrc.json`

Rules for `.sol` files
`.solhint.json` && override section in `prettierrc.json`

package.json
```diff
@@ -44,11 +44,7 @@
     "compile:types": "typechain --target=truffle --outDir build/types/truffle-types \"build/contracts/*.json\"",
     "deploy:development": "npm run build && npm run migrate -- --network development",
     "deploy:kovan": "npm run build && npm run migrate -- --network kovan",
-    "compile:contracts": "truffle compile",
-    "compile:types": "typechain --target=truffle --outDir build/types/truffle-types \"build/contracts/*.json\"",
-    "build": "npm run compile:contracts && npm run compile:types",
     "pretest": "npm run build",
```
haythem96, line 47: Should we remove this `pretest` command ? it just run the `build`

package.json
```diff
@@ -44,11 +44,7 @@
...
-    "test": "truffle test",
     "test:kovan": "truffle test --network kovan",
```
haythem96, line 48: I think also we will not need `test:kovan`, unit tests will be executed locally, right ?

package.json
antoncoding: Okay to me

.prettierrc.json
```diff
@@ -0,0 +1,16 @@
...
+  "overrides": [
+    {
+      "files": "contracts/**/*.sol",
+      "options": {
+        "tabWidth": 4,
+        "singleQuote": false
+      }
+    }
+  ]
+}
```
aparnakr, line 16: does this auto adjust the code or is this more a check?

antoncoding: The default part (for ts) is included in the eslint as an extension, so we will run auto lint in `lint-stage` with `eslint --fix`

The overrides part for .sol files will also be run in `lint-stage` with `prettier --write`

```json
# package.json
"lint-staged": {
    "**/*.sol": [
      "prettier --write",
      "git add"
    ],
    "test/**": [
      "eslint --fix",
      "git add"
    ]
  },
```

PR #6, opened 2020-07-04
OTokenFactory Module

antoncoding:
# PR Overview

Edited: Structure setup related commits are moved to #11. 

## Otoken Factory Overview
* `OtokenFactory` is the contract that use eip-1167 standard to **clone** Otoken contracts. It inherits a [Spawner contract](https://github.com/0age/Spawner/blob/master/contracts/Spawner.sol) that already implemented eip-1167 with CREATE2 opcode (so we have deterministic address).
* The `Spawner` contract is editted so need to be reviewed as well. Some relevant commits are f098951 and 955b5e1
* Before calling `createOtoken()`, user can call `getTargetOtokenAddress`with the same parameter to get the expected address the new cloned otoken contract will be deployed at. 
* The factory uses the hash of `(underlying, strike, collateral, expiry, strikePrice, isPut)` as an id to keep track of the addresses all the options.
  * Changes of `name` and `symbol` will affect the target token address but not the id, the factory won't allow a new otoken to be created if the id already existed.
* User can call `getOtoken(underlying, strike, collateral, expiry, strikePrice, isPut)` to see if the desired series is deployed or not.

## Todos
- [x] High level test cases written out
- [x] NatSpec for each function.
- [x] Implementation
- [x] Make test coverage 100% if not already

### Naming:
1. Better naming for `getTargetOtokenAddress`
2. Better naming for internal variable `_tokenAddresses`


contracts/OtokenFactory.sol
```diff
@@ -0,0 +1,127 @@
+pragma solidity =0.6.0;
+
+import "./lib/Spawner.sol";
+import "./Otoken.sol";
+
+contract OtokenFactory is Spawner {
+    Otoken public logic;
+
+    address[] public otokens;
+    uint256 public otokensCreated;
```
haythem96, line 10: what we need this variable for ?

haythem96: `otokensCreated` is similar to `otokensCreated.length`, right ?

antoncoding: Yes but you cannot get otokens.length off-chain

antoncoding: That's why I think we should just make otokens private, and write our own public getter, that way we can just return the whole array, and we don't need otokensCreated.

haythem96: Agree. Or we can just make a function that return the length of the array, with `otoken` public, and loop through the array off-chain to get all of it, similar to the current implementation in OptionsProtocol repo, dev branch.

But either way could work (y)

antoncoding: I prefer the a public function that return the whole array at once. @aparnakr any thoughts?

.circleci/config.yml
```diff
@@ -2,8 +2,13 @@ version: 2.1
 orbs:
   coveralls: coveralls/coveralls@1.0.6
   node: circleci/node@1.1.6
+workflows:
```
aparnakr, line 5: can you have this in a different PR?

aparnakr: Want to separate the oTokenFactory changes from the circle changes

.gitignore
```diff
@@ -1,2 +1,4 @@
 node_modules/
 build/
+coverage
+coverage.json
```
aparnakr, line 4: this also in a new PR :)

.solcover.js
```diff
@@ -1,16 +1,9 @@
...
-    ]
-};
+  client: require('ganache-cli'),
+  providerOptions: {
+    mnemonic: "candy maple cake sugar pudding cream honey rich smooth crumble sweet treat"
+  },
+  skipFiles: [
+    'Migrations.sol'
+  ]
+}
```
aparnakr, line 9: this also in a new PR :)

README.md
```diff
@@ -1,4 +1,4 @@
-# Gamma Protocol
+# Gamma Protocol [![CircleCI](https://circleci.com/gh/opynfinance/GammaProtocol.svg?style=svg)](https://circleci.com/gh/opynfinance/GammaProtocol/tree/master) [![Coverage Status](https://coveralls.io/repos/github/opynfinance/GammaProtocol/badge.svg?branch=master)](https://coveralls.io/github/opynfinance/GammaProtocol?branch=master)
 A Decentralized Options Protocol built on Ethereum. It supports capital efficient put and call options.
```
aparnakr, line 2: separate PR :)

contracts/Otoken.sol
```diff
@@ -0,0 +1,32 @@
+pragma solidity =0.6.0;
+
+contract Otoken {
```
aparnakr, line 3: Otoken should be ERC20Detailed, that way the name, symbol etc are all taken care of in that

contracts/Otoken.sol
```diff
@@ -0,0 +1,32 @@
...
+    address public strikeAsset;
+    address public underlyingAsset;
+    address public collateralAsset;
+
+    uint256 public strikePrice;
+    uint256 public expiry;
+
+    string public name;
+    string public symbol;
+    uint256 public decimals;
```
aparnakr, line 13: don't need this if it is an ERC20Detailed

.circleci/config.yml
antoncoding: Yeah Haythem added this just to test the code coverage. Let me drop these commits.

antoncoding:
@haythem96 @aparnakr  I pumped the solidity version to 0.6.10 in this PR because the template from openzeppelin requires 0.6.2. 
Also, starts from 0.6.8 the compiler check for license in each file, so I added UNLICENSED in each contract.

contracts/Otoken.sol
antoncoding: I use `ERC20UpgradeSafe` as suggested in this post: [Writing Upgradeable Contracts](https://docs.openzeppelin.com/upgrades/2.8/writing-upgradeable) because we need the ERC20 to have an init function instead of a constructor.

haythem96:
@antoncoding can we use `packages` for third-party contracts and `libs` for libraries ?

antoncoding:
> @antoncoding can we use `packages` for third-party contracts and `libs` for libraries ?

Good idea.

antoncoding:
## Updates:
Add error code system as @aparnakr suggested.
* Add `ErrorReporter.sol` and return error code instead of revert in `createOtoken()`.

Reference: [Compound's ErrorReporter](https://github.com/compound-finance/compound-protocol/blob/master/contracts/ErrorReporter.sol)

aparnakr:
> ## Updates:
> Add error code system as @aparnakr suggested.
> 
> * Add `ErrorReporter.sol` and return error code instead of revert in `createOtoken()`.
> 
> Reference: [Compound's ErrorReporter](https://github.com/compound-finance/compound-protocol/blob/master/contracts/ErrorReporter.sol)

Hey @antoncoding can you move the error coding to a different PR? This is becoming too big

aparnakr:
Guess we should remove the error reporter stuff for now?

contracts/packages/Spawner.sol
```diff
@@ -93,24 +93,18 @@ contract Spawner {
...
-                )
+        bytes32 _data = keccak256( // compute CREATE2 hash using 4 inputs.
+            abi.encodePacked(
+                bytes1(0xff), // pass in the control character.
+                address(this), // pass in the address of this contract.
+                SALT, // pass in the salt from above.
+                initCodeHash // pass in hash of contract creation code.
             )
         );
+        target = address(uint256(_data));
```
haythem96, line 107: In simplifying _computeAddress, there is a missing uint160 cast, can u explain why u removed it ?

antoncoding:
> Guess we should remove the error reporter stuff for now?

Already removed!

contracts/packages/Spawner.sol
antoncoding: OpenZeppelin has a [CREATE2 library](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/Create2.sol), they use this cast and there's a small discussion 
https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2244 on this topic. The `address()` cast will take care of the uint256 => uint160.

contracts/OtokenFactory.sol
```diff
@@ -0,0 +1,177 @@
...
+ * SPDX-License-Identifier: UNLICENSED
+ * @title A factory for opyn tokens
+ * @author Opyn
+ * @notice Create new otokens and keep track of all created tokens.
+ * @dev Calculate contract address before each creation with CREATE2
+ * and deploy eip-1167 minimal proxies for otoken logic contract.
+ */
+contract OtokenFactory is Spawner {
+    // The Opyn AddressBook contract that records addresses of whitelist module and oToken impl contract.
+    IAddressBook public addressBook;
```
aparnakr, line 17: Can we rename to AddressBookInterface? easier to remember it is an interface

contracts/OtokenFactory.sol
```diff
@@ -0,0 +1,177 @@
+pragma solidity =0.6.10;
+
+import {Spawner} from "./packages/Spawner.sol";
+import {IAddressBook} from "./interfaces/IAddressBook.sol";
+import {IWhitelistModule} from "./interfaces/IWhitelistModule.sol";
```
aparnakr, line 5: can we rename these to WhitlelistModuleInterface and AddressBookInterface ?

contracts/OtokenFactory.sol
antoncoding: Are you suggesting changing the variable name `addressbook` or the type `IAddressBook`?

If you're talking about the variable, I don't think it's necessary, because all the external contract instance are interfaces. Also, most contracts name their interfaces this way:  [dydx](https://github.com/dydxprotocol/solo/blob/3b6db45cf109fa2cf272843c4192a7d892ad30f4/contracts/protocol/Admin.sol#L95)

If you're talking about the `IAddressBook` type, I don't have strong opinion on this. I know most people just put `I` to represent an interface, so it's kind of a convention. But Compound always uses `Interface`. 

@haythem96  You have any preference?

haythem96: I mostly use `I` at the beginning, feel like adding the `Interface` make it more longer, but yeah it is okay if u guys wanna go with adding the `Interface`.

contracts/interfaces/IAddressBook.sol
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity 0.6.10;
+
+import {IWhitelistModule} from "./IWhitelistModule.sol";
+import {IOtoken} from "./IOtoken.sol";
+
+interface IAddressBook {
+    function getOtokenImpl() external view returns (IOtoken otoken);
```
haythem96, line 8: shouldn't this one return an address?

haythem96: will be better to return address so the AddressBook module have no dependencies to other modules.

contracts/OtokenFactory.sol
```diff
@@ -0,0 +1,177 @@
...
+
+    mapping(bytes32 => address) private _tokenAddresses;
+
+    constructor(IAddressBook _addressBook) public {
+        addressBook = _addressBook;
+    }
+
+    event OtokenCreated(
+        address tokenAddress,
+        address creator,
```
haythem96, line 29: can we add `indexed` here ?

haythem96:
BTW, for testing, can we use https://www.chaijs.com/api/assert/ instead of 'expect' API ?

contracts/interfaces/IAddressBook.sol
antoncoding: It depends on our decision on the interface coding style, but I'll change everything to `address` type for now.

contracts/OtokenFactory.sol
antoncoding: Updated. we should add event definition to our master doc as well.

antoncoding:
> BTW, for testing, can we use https://www.chaijs.com/api/assert/ instead of 'expect' API ?

Yeah that's a good idea, the syntax is much cleaner.

contracts/OtokenFactory.sol
aparnakr: Yes, but Interface makes it more clear to even a newer solidit developer

contracts/OtokenFactory.sol
```diff
@@ -0,0 +1,177 @@
...
+ * @dev Calculate contract address before each creation with CREATE2
+ * and deploy eip-1167 minimal proxies for otoken logic contract.
+ */
+contract OtokenFactory is Spawner {
+    // The Opyn AddressBook contract that records addresses of whitelist module and oToken impl contract.
+    IAddressBook public addressBook;
+
+    address[] private _otokens;
+
+    mapping(bytes32 => address) private _tokenAddresses;
```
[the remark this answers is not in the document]
aparnakr: This naming is confusing so without comments it makes it even more confusing.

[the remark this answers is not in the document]
antoncoding: it's not product id, so probably just `_idToAddress`

contracts/OtokenFactory.sol
antoncoding: I'm gonna rename to WhitelistInterface. (The term Module doesn't really add anything and it will become too long. )

contracts/OtokenFactory.sol
```diff
@@ -0,0 +1,177 @@
...
+     * @notice create new oTokens
+     * @dev deploy an eip-1167 minimal proxy with CREATE2 and register it to the whitelist module.
+     * @param _underlyingAsset underlying asset
+     * @param _strikeAsset strike asset
+     * @param _collateralAsset collateral asset
+     * @param _strikePrice strike price with decimals = 18
+     * @param _expiry expiration timestamp in second
+     * @param _isPut is this a put option or not
+     * @return newOtoken address of the newly created option
+     */
```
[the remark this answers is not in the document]
aleone:
```suggestion
    /**
     * @param _underlyingAsset asset that the option references
     * @param _strikeAsset asset that the strike price is denominated in
     * @param _collateralAsset asset that is held as collateral against short/written options
     * @param _strikePrice strike price with decimals = 18
     * @param _isPut is this a put option, if not it is a call
     */
```

contracts/interfaces/AddressBookInterface.sol
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity 0.6.10;
+
+interface AddressBookInterface {
+    function getOtokenImpl() external view returns (address otoken);
+
+    function getWhitelist() external view returns (address whitelist);
```
aparnakr, line 7: question on this @haythem96 for return args should we do (type name) or just (type) ? Also how should the name be formatted?

contracts/interfaces/WhitelistInterface.sol
```diff
@@ -0,0 +1,24 @@
...
+        address underlying,
+        address strike,
+        address collateral
+    ) external view returns (bool);
+
+    function isWhitelistedCollateral(address collateral) external view returns (bool);
+
+    function registerOtoken(address oToken) external;
+
+    function isValidOtoken(address oToken) external view returns (bool isValid);
```
aparnakr, line 23: name format for args

contracts/mocks/MockWhitelistModule.sol
```diff
@@ -0,0 +1,30 @@
...
+    function isWhitelistedProduct(
+        address _underlying,
+        address _strike,
+        address _collateral
+    ) external view returns (bool isValid) {
+        bytes32 id = keccak256(abi.encodePacked(_underlying, _strike, _collateral));
+        return _isWhitelistedProduct[id];
+    }
+
+    function registerOtoken(address oToken) external {
```
aparnakr, line 27:
```suggestion
    function registerOtoken(address _oToken) external {
```

contracts/interfaces/AddressBookInterface.sol
antoncoding: I'm okay with either in interfaces. But in actual functions, I like having names for the return variables, it helps a lot in terms of readability when you return more than 1 parameters.

contracts/OtokenFactory.sol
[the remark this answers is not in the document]
antoncoding: Thanks @aleone this is super helpful! Updated.

contracts/interfaces/AddressBookInterface.sol
haythem96: Agree with Anton.

contracts/OtokenFactory.sol
```diff
@@ -0,0 +1,185 @@
...
+ * @notice Create new otokens and keep track of all created tokens.
+ * @dev Calculate contract address before each creation with CREATE2
+ * and deploy eip-1167 minimal proxies for otoken logic contract.
+ */
+contract OtokenFactory is OtokenSpawner {
+    /// @notice The Opyn AddressBook contract that records addresses of whitelist module and otoken impl address. */
+    address public addressBook;
+
+    /// @dev An array of all created otokens */
+    address[] private _otokens;
```
aparnakr, line 21: @haythem96 are we for variables using _name convention or only for functions?

contracts/OtokenFactory.sol
```diff
@@ -0,0 +1,185 @@
...
+
+    /// @dev A mapping from parameters hash to its deployed address
+    mapping(bytes32 => address) private _idToAddress;
```
aparnakr, line 24: same as above

contracts/OtokenFactory.sol
```diff
@@ -0,0 +1,185 @@
...
+
+    constructor(address _addressBook) public {
+        addressBook = _addressBook;
+    }
+
+    /* Event: OtokenCreated - Emitted when factory create a new Option. */
```
aparnakr, line 30:
```suggestion
     /// @notice emitted when factory create a new Option.
```

contracts/OtokenFactory.sol
```diff
@@ -0,0 +1,185 @@
...
+     * @return newOtoken address of the newly created option
+     */
+    function createOtoken(
+        address _underlyingAsset,
+        address _strikeAsset,
+        address _collateralAsset,
+        uint256 _strikePrice,
+        uint256 _expiry,
+        bool _isPut
+    ) external returns (address newOtoken) {
```
aparnakr, line 60: do we need a return address?

contracts/OtokenFactory.sol
```diff
@@ -0,0 +1,185 @@
...
+        require(_expiry > now, "OtokenFactory: Can't create expired option.");
```
aparnakr, line 61: might also want to check expiry is a valid unix timestamp

contracts/OtokenFactory.sol
```diff
@@ -0,0 +1,185 @@
...
+        bytes32 id = _getOptionId(_underlyingAsset, _strikeAsset, _collateralAsset, _strikePrice, _expiry, _isPut);
+        require(_idToAddress[id] == address(0), "OtokenFactory: Option created");
```
aparnakr, line 63:
```suggestion
        require(_idToAddress[id] == address(0), "OtokenFactory: Option already created");
```

contracts/OtokenFactory.sol
```diff
@@ -0,0 +1,185 @@
...
+            OtokenInterface(otokenImpl).init.selector,
+            _underlyingAsset,
+            _strikeAsset,
+            _collateralAsset,
+            _strikePrice,
+            _expiry,
+            _isPut
+        );
+
+        newOtoken = _spawn(AddressBookInterface(addressBook).getOtokenImpl(), initializationCalldata);
```
aparnakr, line 83:
```suggestion
        newOtoken = _spawn(otokenImpl, initializationCalldata);
```

contracts/OtokenFactory.sol
aparnakr: Also do we know how many oTokens are created? Like can we access the length of the array externally? (do we need that data?

contracts/OtokenFactory.sol
aparnakr:
```suggestion
    ) external {
```

contracts/OtokenFactory.sol
aparnakr, line 63: for these two, can you call the getOToken function and make that a public function? When should we make functions external vs public? @haythem96

contracts/OtokenFactory.sol
```diff
@@ -0,0 +1,185 @@
...
+        uint256 _strikePrice,
+        uint256 _expiry,
+        bool _isPut
+    ) external view returns (address otoken) {
+        bytes32 id = _getOptionId(_underlyingAsset, _strikeAsset, _collateralAsset, _strikePrice, _expiry, _isPut);
+        return _idToAddress[id];
+    }
+
+    /**
+     * @notice get the address otoken if call createOtoken with these paramters.
```
aparnakr, line 132:
```suggestion
     * @notice get the address at which a new otoken with these paramters will be deployed
```

contracts/OtokenFactory.sol
```diff
@@ -0,0 +1,185 @@
...
+            _collateralAsset,
+            _strikePrice,
+            _expiry,
+            _isPut
+        );
+        targetAddress = _computeAddress(AddressBookInterface(addressBook).getOtokenImpl(), initializationCalldata);
+    }
+
+    /**
+     * @dev internal function to hash paramters and get option id.
```
aparnakr, line 164:
```suggestion
     * @dev internal function to hash paramters and get option id. Each option has a unique id. 
```

contracts/OtokenSpawner.sol
```diff
@@ -0,0 +1,65 @@
...
+     * given parameters.
+     */
+    function _computeAddress(address logicContract, bytes memory initializationCalldata)
+        internal
+        view
+        returns (address target)
+    {
+        // place creation code and constructor args of contract to spawn in memory.
+        bytes memory initCode = abi.encodePacked(
+            type(Spawn).creationCode,
```
aparnakr, line 57: is there a type(Spawn) from OZ that can be used here?

test/OtokenFactory.test.ts
```diff
@@ -0,0 +1,221 @@
...
+const ZERO_ADDR = '0x0000000000000000000000000000000000000000'
+
+contract('OTokenFactory', accounts => {
+  let otoken: MockOtokenInstance
+  let addressBook: MockAddressBookInstance
+  let otokenFactory: OtokenFactoryInstance
+
+  // Paramter used for otoken init(). (Use random addresses as usdc and eth)
+  let usdc: MockERC20Instance
+  let shitcoin: MockERC20Instance
```
aparnakr, line 25: haha I died

test/OtokenFactory.test.ts
```diff
@@ -0,0 +1,221 @@
...
+        usdc.address,
+        strikePrice,
+        expiry,
+        isPut,
+      )
+      assert.equal(existAddress, ZERO_ADDR, 'getOtoken check failed on undeployed tokens.')
+    })
+
+    it('should get deterministic address with new otoken paramters', async () => {
+      const targetAddress = await otokenFactory.getTargetOtokenAddress(
```
aparnakr, line 68: can you also test that calling this multiple times should give the same address all times?

test/OtokenFactory.test.ts
```diff
@@ -0,0 +1,221 @@
...
+        ZERO_ADDR,
+        strikePrice,
+        expiry,
+        isPut,
+      )
+      assert.notEqual(targetAddress1, targetAddress2)
+    })
+  })
+
+  describe('Create new otoken', () => {
```
aparnakr, line 100: can you test that anyone can create these? Ensure two different addresses can create a valid oToken?

test/OtokenFactory.test.ts
```diff
@@ -0,0 +1,221 @@
...
+        isPut,
+      )
+      assert.equal(existAddress, otoken.address)
+    })
+  })
+
+  describe('Wrong setup: wrong implementation contract', () => {
+    it('Should revert on token creation', async () => {
+      // Set the otoken Impl contract to a wrong address
+      await addressBook.setOtokenImpl(otokenFactory.address)
```
aparnakr, line 212: does this test that only oTokenlogic contracts can be created?

contracts/OtokenFactory.sol
antoncoding: Actually I'm thinking about can we just have a counter and a public mapping. Returning the whole array could be problematic when we have lots of otokens in the future. (Similar to the `getVaultOwners` function we have in v1...)

haythem96: We can have a public array and a counter, and a getter function that return the counter, so off-chain it just need to loop from 0 to counter-1, similar to here https://github.com/opynfinance/OptionsProtocol/blob/master/contracts/OptionsFactory.sol#L102

contracts/OtokenFactory.sol
antoncoding: I think it would be helpful in terms of contract integration. 

For example for the following flow:
```
address otoken = Factory.getOtoken(params...)
if(otoken = address(0) ) {
   otoken = Factory.createOtoken(params...)
} 
```

contracts/OtokenFactory.sol
haythem96: Yeah those two lines can be refactored by calling the `getOtoken`.
In this case L62 and `_idToAddress[id] ` are replicating the getOtoken function, so it make sense to make it public and call it.

contracts/OtokenSpawner.sol
antoncoding: No, but the `Create2.sol` library by oz does point out this is the correct way to get the contract creation code:

```
library Create2 {
    /**
     * @dev Deploys a contract using `CREATE2`. The address where the contract
     * will be deployed can be known in advance via {computeAddress}.
     *
     * The bytecode for a contract can be obtained from Solidity with
     * `type(contractName).creationCode`.
     *
```

test/OtokenFactory.test.ts
antoncoding: yeah, it tests that the creation will fail if the address book has the wrong `otokenImpl` address

contracts/OtokenFactory.sol
haythem96: I don't think that's a good idea, we should make that check in Factory.
In case `otoken == address(0)` there will be two calls to the Factory, which is more expensive in term of gas.

contracts/OtokenFactory.sol
antoncoding: @aleone and I were chatting about potentially add more restriction on expiry, but not sure should we have them here or in the Otoken contract
I currently plan to have the max expiry check in the `oToken` init function (the expiry can't be greator than the max uint bokkypoobah tested ), because Otoken contract use the bokkypoobah library.  Should we move that check here as well so we can be more consistent?

Another check to include: 
* (expiry - 28800) mod 86400 == 0, to ensure all the option expires on 08:00 UTC.

@aparnakr  @haythem96

contracts/OtokenFactory.sol
antoncoding: I use the `id` here because in the later part of the function, I need to set
```
_idToAddress[id] = newOtoken;
```

I don't want to calculate the id twice.

haythem96: mm yeah make sense (y).

test/OtokenFactory.test.ts
antoncoding: added

test/OtokenFactory.test.ts
antoncoding: it's in `Create2.sol` 
```
require(addr != address(0), "Create2: Failed on deploy");        
```

I'm gonna update the revert message to this one

contracts/OtokenFactory.sol
antoncoding: @haythem96 that's a good idea, it's how Uniswap v2 handles it too.

https://github.com/Uniswap/uniswap-v2-core/blob/4dd59067c76dea4a0e8e4bfdda41877a6b16dedc/contracts/UniswapV2Factory.sol#L19

contracts/OtokenFactory.sol
antoncoding: You're right we don't need `_` for private variables.

contracts/OtokenFactory.sol
antoncoding: > 1. Will you have integration tests to test how this works with whitelist module?

We will need addressBook ready to run real integration tests on all of them

antoncoding:
> Questions:
> 1. Will you have integration tests to test how this works with whitelist module?

@aparnakr we will need AddressBook ready to run integration tests.

contracts/OtokenFactory.sol
antoncoding: I included following 2 checks in 804b7ea
*  all expiry has to be 08:00 UTC
* expiry < 2345/12/31, which is the max DatetimeLibrary supports

contracts/OtokenFactory.sol
aparnakr: hmm it breaks our syntax of not having returns in external functions, but if it is helpful in this case, that's ok imo. @haythem96 thoguths?

contracts/OtokenFactory.sol
aparnakr: hmm does it need to be 08:00 UTC? Will we have expiries that are different @aleone?

PR #11, opened 2020-07-14
Contract folder setup 

antoncoding:
# PR Overview

The setup part is moved here from #6  

## Folder structure overview:
`contracts/interfaces`: Include all the interfaces for our contracts, which should all start with capital `I`.
`contracts/packages`: [updated] All the third party contracts/ libraries that are inherited or imported by our contracts should live in this folder.
`contracts/packages/oz`: [updated] All the contracts from open-zeppelin contracts package.
`contracts/libs`: Our own libraries.
`contracts/test/`: Where all the mock contracts written for testing lives.
### Other setups
* Compiler version pump: Upgrade from 0.6.0 => 0.6.10 since OZ's ERC20 implementation needs 0.6.2.
* Little update to `math.test.ts` so we have 100% coverage on master initially.


haythem96:
looks good! can we change `contracts/test/` to `contracts/mocks/` ?

PR #12, opened 2020-07-17
Code structure for Whitelist Module + oToken

aparnakr:


contracts/WhitelistModule.sol
```diff
@@ -0,0 +1,94 @@
...
+    /**
+     * @notice TODO:
+     * @dev TODO:
+     * @param collateral TODO:
+     */
+    function whitelistCollateral(address collateral) external onlyOwner {
+        /** TODO: */
+    }
+
+    modifier onlyFactory() {
```
antoncoding, line 81: I think we don't need this, because only the `init` function is called by the factory, and it already has a "Initializer" modifier, so it can only be called once.

We can discuss more during implementation.

package.json
```diff
@@ -55,9 +55,9 @@
   },
   "lint-staged": {
     "**/*.sol": [
-      "npm run lint:sol",
       "npm run lint:sol:prettier:fix",
-      "git add"
+      "git add",
+      "npm run lint:sol"
```
antoncoding, line 60: Why moving `npm run lint:sol` to the last, wouldn't this result in still stage files (`git add`) if there's a lint error?

contracts/Otoken.sol
```diff
@@ -0,0 +1,68 @@
...
+     * @param expiryTimestamp TODO:
+     * @param strikePrice TODO:
+     */
+    function init(
+        address underlyingAsset,
+        address strikeAsset,
+        address collateralAsset,
+        bool isPut,
+        uint256 expiryTimestamp,
+        uint256 strikePrice
```
antoncoding, line 64: How do people feel about the order of the parameters? I think we should probably somehow follow our naming
`ETHUSDC/{timestamp}/205P/USDC`, so I think it's better to use: `underlyingAsset, strikeAsset, collateralAsset, expiryTimestamp, strikePrice, isPut`

PR #13, opened 2020-07-18
Project structure

haythem96:
This PR add the basic project contracts files.

### DIFF
--- a/contracts/Controller.sol
+++ b/contracts/Controller.sol
@@ -1,0 +1,14 @@
+/**
+ * SPDX-License-Identifier: UNLICENSED
+ */
+pragma solidity 0.6.10;
+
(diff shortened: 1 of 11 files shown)


contracts/Controller.sol
```diff
@@ -0,0 +1,14 @@
...
+ */
+pragma solidity 0.6.10;
+
+import {MarginCalculator} from "./MarginCalculator.sol";
+
+/**
+ *
+ */
+// solhint-disable-next-line no-empty-blocks
+contract Controller is MarginCalculator {
```
antoncoding, line 12: Aren't we separating these two into two contracts?

haythem96: Ah wasn't sure if we doing that or we writing everything in the same contract.

PR #19, opened 2020-07-20
Whitelist otoken

haythem96:
This PR add the feature of whitelisting a specific `otoken` by the `Otoken Factory`, related to the `Whitelist` Module.

### DIFF
--- a/contracts/Whitelist.sol
+++ b/contracts/Whitelist.sol
@@ -2,7 +2,6 @@
 pragma solidity 0.6.10;
 
 import "./interfaces/AddressBookInterface.sol";
-
 import "./packages/oz/Ownable.sol";
 
 /**
@@ -14,9 +13,9 @@ contract Whitelist is Ownable {
     ///@notice AddressBook module address
     address public addressBook;
 
-    ///@dev mapping to track whitelisted product
+    ///@dev mapping to track whitelisted products
     mapping(bytes32 => bool) internal whitelistedProduct;
-    ///@dev mapping to track whitelisted otoken
+    ///@dev mapping to track whitelisted otokens
     mapping(address => bool) internal whitelistedOtoken;
 
     /**
@@ -29,7 +28,7 @@ contract Whitelist is Ownable {
         addressBook = _addressBook;
     }
 
-    ///@notice emitted when owner whitelist a product
+    ///@notice emitted when the owner whitelists a new product
     event ProductWhitelisted(
         bytes32 productHash,
         address indexed underlying,
--- a/contracts/Whitelist.sol
+++ b/contracts/Whitelist.sol
@@ -39,7 +39,7 @@ contract Whitelist is Ownable {
     }
 
     /**
-     * @notice whitelist a product
+     * @notice allow owner to whitelist product
      * @dev a product is the hash of the underlying, collateral and strike assets
      * can only be called from owner address
      * @param _underlying option underlying asset address
--- a/contracts/Whitelist.sol
+++ b/contracts/Whitelist.sol
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: UNLICENSED
 pragma solidity 0.6.10;
 
+import "./interfaces/AddressBookInterface.sol";
+
 import "./packages/oz/Ownable.sol";
 
 /**
@@ -9,8 +11,23 @@ import "./packages/oz/Ownable.sol";
  * @notice The whitelist module keeps track of all valid Otoken contracts.
  */
 contract Whitelist is Ownable {
-    ///@notice mapping to track whitelisted product
+    ///@notice AddressBook module address
+    address public addressBook;
+
+    ///@dev mapping to track whitelisted product
     mapping(bytes32 => bool) internal whitelistedProduct;
+    ///@dev mapping to track whitelisted otoken
+    mapping(address => bool) internal whitelistedOtoken;
+
+    /**
+     * @dev constructor
+     * @param _addressBook AddressBook module address
+     */
+    constructor(address _addressBook) public {
+        require(_addressBook != address(0), "Invalid address book");
+
+        addressBook = _addressBook;
+    }
 
     ///@notice emitted when owner whitelist a product
     event ProductWhitelisted(
@@ -20,13 +37,25 @@ contract Whitelist is Ownable {
         address indexed collateral
     );
 
+    ///@notice emitted when Otoken Factory module whitelist an otoken
+    event OtokenWhitelisted(address otoken);
+
     /**
-     * @notice check if a product is supported
+     * @notice check if sender is Otoken Factory module
+     */
+    modifier onlyFactory() {
+        require(msg.sender == AddressBookInterface(addressBook).getOtokenFactory(), "Sender is not Otoken Factory");
+
+        _;
+    }
+
+    /**
+     * @notice check if a product is whitelisted
      * @dev product = the hash of underlying, strike and collateral asset
      * @param _underlying option underlying asset address
      * @param _strike option strike asset address
      * @param _collateral option collateral asset address
-     * @return boolean, true if product is supported
+     * @return boolean, true if product is whitelisted
      */
     function isWhitelistedProduct(
         address _underlying,
@@ -36,6 +65,15 @@ contract Whitelist is Ownable {
         bytes32 productHash = keccak256(abi.encode(_underlying, _strike, _collateral));
 
         return whitelistedProduct[productHash];
+    }
+
+    /**
+     * @notice check if an otoken is whitelisted
+     * @param _otoken otoken address
+     * @return boolean, true if otoken is whitelisted
+     */
+    function isWhitelistedOtoken(address _otoken) external view returns (bool) {
+        return whitelistedOtoken[_otoken];
     }
 
     /**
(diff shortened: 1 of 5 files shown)


contracts/Whitelist.sol
```diff
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: UNLICENSED
 pragma solidity 0.6.10;
 
+import "./interfaces/AddressBookInterface.sol";
+
```
aparnakr, line 5: delete the extra empty line

contracts/Whitelist.sol
```diff
@@ -9,8 +11,23 @@ import "./packages/oz/Ownable.sol";
...
+    /**
+     * @dev constructor
+     * @param _addressBook AddressBook module address
+     */
+    constructor(address _addressBook) public {
+        require(_addressBook != address(0), "Invalid address book");
+
+        addressBook = _addressBook;
+    }
```
aparnakr, line 30: Can anyone call the constructor or just the owner?

contracts/Whitelist.sol
```diff
@@ -20,13 +37,25 @@ contract Whitelist is Ownable {
         address indexed collateral
     );
 
+    ///@notice emitted when Otoken Factory module whitelist an otoken
```
aparnakr, line 40: `///@notice emitted when the Otoken Factory module whitelists an otoken`

contracts/Whitelist.sol
```diff
@@ -9,8 +11,23 @@ import "./packages/oz/Ownable.sol";
...
+     * @dev constructor
+     * @param _addressBook AddressBook module address
+     */
+    constructor(address _addressBook) public {
+        require(_addressBook != address(0), "Invalid address book");
+
+        addressBook = _addressBook;
+    }
 
     ///@notice emitted when owner whitelist a product
```
aparnakr, line 32: `///@notice emitted when the owner whitelists a new product`

contracts/Whitelist.sol
```diff
@@ -9,8 +11,23 @@ import "./packages/oz/Ownable.sol";
  * @notice The whitelist module keeps track of all valid Otoken contracts.
  */
 contract Whitelist is Ownable {
-    ///@notice mapping to track whitelisted product
+    ///@notice AddressBook module address
+    address public addressBook;
+
+    ///@dev mapping to track whitelisted product
```
aparnakr, line 17: `///@dev mapping to track whitelisted products`

contracts/Whitelist.sol
```diff
@@ -9,8 +11,23 @@ import "./packages/oz/Ownable.sol";
...
     mapping(bytes32 => bool) internal whitelistedProduct;
+    ///@dev mapping to track whitelisted otoken
```
aparnakr, line 19: `///@dev mapping to track whitelisted otokens`

contracts/Whitelist.sol
haythem96: constructor called only at deployment time.

contracts/Whitelist.sol
antoncoding: so when are we using /// @notice ?

contracts/Whitelist.sol
```diff
@@ -9,24 +10,51 @@ import "./packages/oz/Ownable.sol";
...
 
+    ///@notice emitted when Otoken Factory module whitelist an otoken
+    event OtokenWhitelisted(address otoken);
+
     /**
-     * @notice check if a product is supported
+     * @notice check if sender is Otoken Factory module
+     */
+    modifier onlyFactory() {
+        require(msg.sender == AddressBookInterface(addressBook).getOtokenFactory(), "Sender is not Otoken Factory");
```
antoncoding, line 46: I suggest adding the contract name to the require message, ( `Sender is not Otoken Factory` => `WhiteList: Sender is not Otoken Factory`, `WL:Sender is not Otoken Factory`)  that way it's much easier for us to know which contract is throwing errors while testing.

test/Whitelist.test.ts
```diff
@@ -1,16 +1,21 @@
...
 const {expectEvent, expectRevert} = require('@openzeppelin/test-helpers')
 
 const MockERC20 = artifacts.require('MockERC20.sol')
+const MockAddressBook = artifacts.require('MockAddressBook.sol')
 const Whitelist = artifacts.require('Whitelist.sol')
 // address(0)
 const ZERO_ADDR = '0x0000000000000000000000000000000000000000'
 
-contract('Whitelist', ([owner, random, newOwner]) => {
+contract('Whitelist', ([owner, otokenFactoryAddress, random, newOwner]) => {
```
antoncoding, line 11: nice, we should use this

contracts/Whitelist.sol
```diff
@@ -59,12 +96,33 @@ contract Whitelist is Ownable {
     }
 
     /**
-     * @notice set a product hash as supported
+     * @notice allow Otoken Factory to whitelist option
+     * @dev can only be called from owner address
+     * @param _otokenAddress otoken
+     */
+    function whitelistOtoken(address _otokenAddress) external onlyFactory {
```
antoncoding, line 103: Didn't we change the name to `registerOtoken` already?

contracts/Whitelist.sol
```diff
@@ -38,6 +66,15 @@ contract Whitelist is Ownable {
         return whitelistedProduct[productHash];
     }
 
+    /**
+     * @notice check if an otoken is whitelisted
+     * @param _otoken otoken address
+     * @return boolean, true if otoken is whitelisted
+     */
+    function isWhitelistedOtoken(address _otoken) external view returns (bool) {
+        return whitelistedOtoken[_otoken];
```
antoncoding, line 75: `isValidOtoken` make more sense to me. It's not like we only whitelist several otokens out of all the otokens, we literally whitelist every valid token created by the factory.

contracts/Whitelist.sol
antoncoding: I saw your comments on the interface thread, but I think we better discuss this more.

contracts/Whitelist.sol
haythem96: yeah thought better to use same terminology everywhere, "Whitelist module, whitelist, isWhitelisted", even if we whitelist all token, it is still the same as valid as we are marking all token as valid. I think better to use same terminology.

antoncoding: Let's use this for now, could easily change it back if we end up having different decision.

haythem96:
> Need to resolve the oToken naming part

What naming part ?

contracts/Whitelist.sol
aparnakr: for events and variables!

contracts/Whitelist.sol
```diff
@@ -9,24 +10,54 @@ import "./packages/oz/Ownable.sol";
...
         address indexed strike,
         address indexed collateral
     );
 
+    ///@notice emitted when Otoken Factory module whitelist an otoken
+    event OtokenWhitelisted(address otoken);
+
     /**
-     * @notice check if a product is supported
+     * @notice check if sender is Otoken Factory module
```
aparnakr, line 43:
```suggestion
     * @notice check if the sender is the Otoken Factory module
```

contracts/Whitelist.sol
```diff
@@ -9,24 +10,54 @@ import "./packages/oz/Ownable.sol";
...
+
+        _;
+    }
+
+    /**
+     * @notice check if a product is whitelisted
      * @dev product = the hash of underlying, strike and collateral asset
      * @param _underlying option underlying asset address
      * @param _strike option strike asset address
      * @param _collateral option collateral asset address
```
aparnakr, line 59: update these with the comment @aleone had on https://github.com/opynfinance/GammaProtocol/pull/6

contracts/Whitelist.sol
aparnakr: +1 on Anton's comment

contracts/Whitelist.sol
```diff
@@ -59,12 +99,33 @@ contract Whitelist is Ownable {
     }
 
     /**
-     * @notice set a product hash as supported
+     * @notice allow Otoken Factory to whitelist option
+     * @dev can only be called from owner address
```
aparnakr, line 103:
```suggestion
     * @dev can only be called from the owner's address
```

contracts/Whitelist.sol
```diff
@@ -59,12 +99,33 @@ contract Whitelist is Ownable {
     }
 
     /**
-     * @notice set a product hash as supported
+     * @notice allow Otoken Factory to whitelist option
```
aparnakr, line 102:
```suggestion
     * @notice allow Otoken Factory to whitelist a new option
```

contracts/Whitelist.sol
```diff
@@ -59,12 +99,33 @@ contract Whitelist is Ownable {
...
      */
     function _setWhitelistedProduct(bytes32 _productHash) internal {
         require(whitelistedProduct[_productHash] == false, "Product already whitelisted");
 
         whitelistedProduct[_productHash] = true;
     }
+
+    /**
+     * @notice set an otoken address as whitelisted
+     * @param _otokenAddress product hash in bytes
```
aparnakr, line 126:
```suggestion
     * @param _otokenAddress address of the oToken that is being whitelisted
```

test/Whitelist.test.ts
```diff
@@ -50,6 +63,28 @@ contract('Whitelist', ([owner, random, newOwner]) => {
     })
   })
 
+  describe('Whitelist otoken', () => {
+    it('should revert whitelisting an otoken from sender other than Otoken Factory', async () => {
+      await expectRevert(whitelist.whitelistOtoken(otoken.address, {from: owner}), 'Sender is not Otoken Factory')
+    })
+
+    it('should whitelist an otoken from Otoken Factory address', async () => {
```
aparnakr, line 71: Can we add a test to ensure that after Factory Address has changed in the addressBook, the new oToken can still be whitelisted

aparnakr:
Great work!

test/Whitelist.test.ts
haythem96: In upgradability, the first deployed address will alway be the caller, as `V1` will delegate call to `V2`, then `V2` will execute, in the execution transaction, `V1` is the `msg.sender`. (V1 is first version of the contract and V2 is an upgrade to a second version).

haythem96: Or no I think I'm wrong in this, will make sure of it, if the contract address change when it make an external call, will add the needed test for that.

haythem96: Sorry my first response is out-of-topic, so when we change the implementation address of the Factory module in the addressBook, nothing change for the whitelist module as the `onlyFactory` modifier always call the addressBook to get the updated implementation address.

PR #20, opened 2020-07-20
Update git hook.

haythem96:
This PR remove unnecessary command from git hook. Basically `npm run lint:sol:prettier:fix` will fix linting error, but that command will never get executed as the hook fail if there are linting errors.

Will be better to just execute that command manually to fix linting, as it sometime move lines/indentations in an ugly way.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,6 @@
   "lint-staged": {
     "**/*.sol": [
       "npm run lint:sol",
-      "npm run lint:sol:prettier:fix",
       "git add"
     ],
     "test/**": [


package.json
```diff
@@ -57,7 +57,6 @@
   "lint-staged": {
     "**/*.sol": [
       "npm run lint:sol",
-      "npm run lint:sol:prettier:fix",
```
aparnakr, line 60: wait when should this be run then?

haythem96: when u commit, if it show linting error, run that command to fix errors.

aparnakr: why not just run it every time?

haythem96: if 'npm run lint:sol' fail, that command will not run
if 'npm run lint:sol' succeed, that command will not do anything when it run because basically linting is good, right ?

So no need for that command IMO.

PR #24, opened 2020-07-21
Otoken: init function

antoncoding:
### Description

This PR is for the init function for `Otoken` contract. In this function we will 
* set global parameters.
* generate deterministic name and symbol

### Todos:
- [x] Add doc for each function
- [x] Make test caes 100% coverage

### Decisions
- [x] What `name` makes the most sense
  * ETHUSDC 05-September-2020 200 Put USDC Collateral
- [x] What `symbol` makes the most sense
  * oETHUSDC-05SEP20-200P


antoncoding:
The on-chain name and symbol generation is making the `init` function very expensive, now this function is costing 230000 gas (around $3.3 with current 60 gwei gas price.)

image

contracts/Otoken.sol
```diff
@@ -1,12 +1,173 @@
...
+    using SafeMath for uint256;
+    address public addressBook;
+    address public underlyingAsset;
+    address public strikeAsset;
+    address public collateralAsset;
+
+    uint256 public strikePrice;
+    uint256 public expiry;
+
+    bool public isPut;
```
aparnakr, line 25: comments with descriptions for variables!

contracts/Otoken.sol
```diff
@@ -1,12 +1,173 @@
...
+
+    uint256 private constant STRIKE_PRICE_DIGITS = 10**18;
```
aparnakr, line 27: use 1e18 instead of 10**x everywhere :) @haythem96 thoughts on what we should use to standardize?

contracts/Otoken.sol
```diff
@@ -1,12 +1,173 @@
...
+
+    constructor(address _addressBook) public {
+        addressBook = _addressBook;
+    }
+
+    /**
+     * @notice initialize a new otokens.
```
aparnakr, line 34:
```suggestion
     * @notice initialize new otokens
```

contracts/Otoken.sol
```diff
@@ -1,12 +1,173 @@
...
+     * @param _underlyingAsset asset that the option references
+     * @param _strikeAsset asset that the strike price is denominated in
+     * @param _collateralAsset asset that is held as collateral against short/written options
+     * @param _strikePrice strike price with decimals = 18
+     * @param _expiry expiration timestamp in second
```
aparnakr, line 39:
```suggestion
     * @param _expiry expiration time of the option represented by unix timestamp
```

contracts/Otoken.sol
```diff
@@ -1,12 +1,173 @@
...
+            _expiry,
+            _isPut
+        );
+        __ERC20_init_unchained(name, symbol);
+    }
+
+    /**
+     * @notice return option name as ETHUSDC/1597512453/200P/USDC, symbol ETHUSDC/1597512453/200P/USDC
+     * @return name ETH-USDC 200 Put 1597512453
+     * @return symbol ETHUSDC/1597512453/200P/USDC
```
aparnakr, line 70: has this been decided or is this still TBD?

test/Otoken.test.ts
```diff
@@ -0,0 +1,85 @@
...
+      const _expiry = (await otoken.expiry()).toNumber()
+      assert.equal(_underlying, ETH_ADDR)
+      assert.equal(_strike, usdc.address)
+      assert.equal(_collateral, usdc.address)
+      assert.equal(_strikePrice, strikePrice.toString())
+      assert.equal(_isPut, isPut)
+      assert.equal(_expiry, expiry)
+    })
+
+    it('should revert when init is called twice', async () => {
```
aparnakr, line 78:
```suggestion
    it('should revert when init is called twice on the parameters', async () => {
```

test/Otoken.test.ts
```diff
@@ -0,0 +1,85 @@
...
+      await expectRevert(
+        otoken.init(ETH_ADDR, usdc.address, usdc.address, strikePrice, expiry, isPut),
+        'Contract instance has already been initialized.',
```
aparnakr, line 81: should there be a test on the upper limit of the date?

aparnakr: Also is ETH going to be represented in the system as 0x0 or WETH? @haythem96 thoughts?

contracts/Otoken.sol
haythem96: yeah I agree, IMO 1e18 looks better.

test/Otoken.test.ts
haythem96: We can use WETH and wrap from ETH to WETH using https://github.com/gnosis/canonical-weth, or we can think of another hack for the address(0) thing... let's decide this next eng sync!

contracts/Otoken.sol
antoncoding: TBD~

contracts/Otoken.sol
antoncoding: KK

contracts/Otoken.sol
```diff
@@ -1,12 +1,197 @@
...
+    ) external initializer {
+        require(_expiry < 11865398400, "Otoken: Can't init with expiry > 2345/12/31.");
+        underlyingAsset = _underlyingAsset;
+        strikeAsset = _strikeAsset;
+        collateralAsset = _collateralAsset;
+        strikePrice = _strikePrice;
+        expiry = _expiry;
+        isPut = _isPut;
+        (string memory name, string memory symbol) = _getNameAndSymbol();
+        __ERC20_init_unchained(name, symbol);
```
aparnakr, line 71: where do we set the decimals?

test/Otoken.test.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  before('Deployment', async () => {
+    // Need another mock contract for addressbook when we add ERC20 operations.
+    otoken = await Otoken.new(mockAddressBook)
+    usdc = await MockERC20.new('USDC', 'USDC')
+  })
+
+  describe('Otoken Initialization', () => {
+    it('should be able to initialize with put parameter correctly', async () => {
+      await otoken.init(ETH_ADDR, usdc.address, usdc.address, strikePrice, expiry, isPut, {from: deployer})
```
aparnakr, line 28: shouldn't this only be callable from Factory?

antoncoding: we don't need to restrict that in my opinion, we only need to make sure it can only be called once, because our factory will call init once right after the creation, and we can just use the `Initializer` modifier from OZ

contracts/Otoken.sol
antoncoding: In `__ERC20_init_unchained ` it's default to 18

antoncoding:
> Great job on this! Again a lot of complexity going on with the date time library :) You did a good job of distilling that down!
> 
> Also question on will the integration tests with Factory + Whitlist Module be in this or elsewhere?

We need the AddressBook module to do real integration test.

test/Otoken.test.ts
antoncoding: If we really want to reduce complexity, we definitely don't need that one more modifier. 

The onlyFactory feature is automatically satisfied when you combine the Factory and the Otoken, so that should be included in the integration test: Factory ensure it calls init right after deploy the contract, and Otoken make sure the function can only be called once.

So in the unit test for Otoken, we only need to test that it can only be called once.

PR #25, opened 2020-07-21
Whitelist collateral functionality

aleone:
This PR adds code needed for the whitelist collateral functionality.

Added: 

- isWhitelistedCollateral
- whitelistCollateral
- _setWhitelistedCollateral
- associated variables and events


contracts/Whitelist.sol
```diff
@@ -9,17 +9,22 @@ import "./packages/oz/Ownable.sol";
  * @notice The whitelist module keeps track of all valid Otoken contracts.
  */
 contract Whitelist is Ownable {
-    ///@notice mapping to track whitelisted product
+    /// @notice mapping to track whitelisted product
```
antoncoding, line 12: Does your editor throw warnings here? We could change this tag from `@notice` to `@dev` 

```
Documentation tag on non-public state variables will be disallowed in 0.7.0. You will need to use the @dev tag explicitly.
```

contracts/Whitelist.sol
```diff
@@ -67,4 +93,14 @@ contract Whitelist is Ownable {
 
         whitelistedProduct[_productHash] = true;
     }
+
+    /**
+     * @notice set a collateral address as whitelisted
+     * @param _collateral collateral address
+     */
+    function _setWhitelistedCollateral(address _collateral) internal {
+        require(whitelistedCollateral[_collateral] == false, "Collateral already whitelisted");
```
antoncoding, line 101: Same suggestion as I give to Haythem, we could include the contract name in the require message to smooth the testing process. (`Collateral already whitelisted ` => `Whitelist: Collateral already whitelisted.`)

contracts/Whitelist.sol
```diff
@@ -9,17 +9,22 @@ import "./packages/oz/Ownable.sol";
  * @notice The whitelist module keeps track of all valid Otoken contracts.
  */
 contract Whitelist is Ownable {
-    ///@notice mapping to track whitelisted product
+    /// @notice mapping to track whitelisted product
     mapping(bytes32 => bool) internal whitelistedProduct;
+    /// @notice mapping to track whitelised collateral;g
```
aparnakr, line 14:
```suggestion
    /// @notice mapping to track whitelised collateral
```

contracts/Whitelist.sol
```diff
@@ -9,17 +9,22 @@ import "./packages/oz/Ownable.sol";
...
-    ///@notice emitted when owner whitelist a product
+    /// @notice emitted when owner whitelist a product
     event ProductWhitelisted(
         bytes32 productHash,
         address indexed underlying,
         address indexed strike,
         address indexed collateral
     );
 
+    /// @notice emits an event when a collateral address is whitelisted by owner address
```
aparnakr, line 25:
```suggestion
    /// @notice emits an event when a collateral address is whitelisted by the owner address
```

contracts/Whitelist.sol
```diff
@@ -39,7 +44,16 @@ contract Whitelist is Ownable {
     }
 
     /**
-     * @notice allow owner to whitelist product
+     * @notice check if the collateral is whitelisted
+     * @param _collateral collateral asset address
```
aparnakr, line 48: update description of _collateral

contracts/Whitelist.sol
```diff
@@ -58,6 +72,17 @@ contract Whitelist is Ownable {
         emit ProductWhitelisted(productHash, _underlying, _strike, _collateral);
     }
 
+    /**
+     * @notice whitelist a collateral address, can only be called by owner
+     * @dev function can only be called by owner
+     * @param _collateral collateral asset address
+     */
+    function whitelistCollateral(address _collateral) external onlyOwner {
+        _setWhitelistedCollateral(_collateral);
```
aparnakr, line 81: are we using ETH as 0x000? or are we using WETH?

aparnakr: Also shouldn't be able to whitelist a collateral that is already whitelisted

test/Whitelist.test.ts
```diff
@@ -50,6 +50,28 @@ contract('Whitelist', ([owner, random, newOwner]) => {
...
+      expectEvent(whitelistTx, 'CollateralWhitelisted')
+
+      const isWhitelistedCollateral = await whitelist.isWhitelistedCollateral(usdc.address)
+      assert.equal(isWhitelistedCollateral, true, 'fail: collateral not whitelisted')
+    })
+
+    it('should revert whitelisting an already whitelisted collateral', async () => {
+      await expectRevert(whitelist.whitelistCollateral(usdc.address, {from: owner}), 'Collateral already whitelisted')
+    })
+  })
```
aparnakr, line 73: would need a test around WETH / ETH once we decide that as well

contracts/Whitelist.sol
haythem96: @aparnakr the checks for an already whitelisted collateral is in `_setWhitelistedCollateral()`.
For address(0) I think we can keep it this way for now and review this part when we decide.

PR #29, opened 2020-07-22
DepositArgs Actions

aparnakr:
This PR adds code needed for the DepositArgs Action functionality

Added:

ActionArgs
DepositArgs
_parseDepositArgs

### DIFF
--- a/contracts/libs/Actions.sol
+++ b/contracts/libs/Actions.sol
@@ -17,6 +17,7 @@ library Actions {
         WithdrawLongOption,
         DepositCollateral,
         WithdrawCollateral,
+        SettleVault,
         Exercise,
         Call
     }
@@ -30,7 +31,7 @@ library Actions {
         address sender;
         /// @notice The asset that is to be transfered
         address asset;
-        /// @notice The address of the vault that is to be modified (if any)
+        /// @notice The index of the vault that is to be modified (if any)
         uint256 vaultId;
         /// @notice The amount of asset that is to be transfered
         uint256 amount;
@@ -46,7 +47,7 @@ library Actions {
     struct DepositArgs {
         /// @notice The address of the account owner
         address owner;
-        /// @notice The address of the vault to which the asset will be added
+        /// @notice The index of the vault to which the asset will be added
         uint256 vaultId;
         /// @notice The address from which we transfer the asset
         address from;
@@ -68,20 +69,19 @@ library Actions {
      */
     function _parseDepositArgs(ActionArgs memory _args) internal returns (DepositArgs memory) {
         require(
-            (uint256(_args.actionType) == uint256(ActionType.DepositLongOption)) ||
-                (uint256(_args.actionType) == uint256(ActionType.DepositCollateral)),
+            (_args.actionType == ActionType.DepositLongOption) || (_args.actionType == ActionType.DepositCollateral),
             "Actions: can only parse arguments for deposit actions"
         );
         require(_args.owner != address(0), "Actions: cannot deposit to an invalid account");
 
-        DepositArgs memory args;
-        args.owner = _args.owner;
-        args.vaultId = _args.vaultId;
-        args.from = _args.sender;
-        args.asset = _args.asset;
-        args.index = _args.index;
-        args.amount = _args.amount;
-
-        return args;
+        return
+            DepositArgs({
+                owner: _args.owner,
+                vaultId: _args.vaultId,
+                from: _args.sender,
+                asset: _args.asset,
+                index: _args.index,
+                amount: _args.amount
+            });
     }
 }
--- a/contracts/libs/Actions.sol
+++ b/contracts/libs/Actions.sol
@@ -60,4 +60,28 @@ library Actions {
         /// @notice The amount of asset that is to be transfered
         uint256 amount;
     }
+
+    /**
+     * @notice Parses the passed in action argmuents to get the argmuents for a deposit action
+     * @param _args The general action arguments structure
+     * @return The arguments for a deposit action
+     */
+    function _parseDepositArgs(ActionArgs memory _args) internal returns (DepositArgs memory) {
+        require(
+            (uint256(_args.actionType) == uint256(ActionType.DepositLongOption)) ||
+                (uint256(_args.actionType) == uint256(ActionType.DepositCollateral)),
+            "Actions: can only parse arguments for deposit actions"
+        );
+        require(_args.owner != address(0), "Actions: cannot deposit to an invalid account");
+
+        DepositArgs memory args;
+        args.owner = _args.owner;
+        args.vaultId = _args.vaultId;
+        args.from = _args.sender;
+        args.asset = _args.asset;
+        args.index = _args.index;
+        args.amount = _args.amount;
+
+        return args;
+    }
 }
--- a/contracts/libs/Actions.sol
+++ b/contracts/libs/Actions.sol
@@ -8,5 +8,56 @@ pragma solidity 0.6.10;
  */
 // solhint-disable-next-line no-empty-blocks
 library Actions {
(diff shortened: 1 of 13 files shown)


aparnakr:
Once this is reviewed, it should be pretty easy to copy paste the remaining actions! :)

contracts/libs/Actions.sol
```diff
@@ -8,5 +8,80 @@ pragma solidity 0.6.10;
...
+    enum ActionType {
+        OpenVault,
+        MintShortOption,
+        BurnShortOption,
+        DepositLongOption,
+        WithdrawLongOption,
+        DepositCollateral,
+        WithdrawCollateral,
+        Exercise,
+        Call
```
antoncoding, line 21: And ClearVault?

contracts/libs/Actions.sol
```diff
@@ -8,5 +8,80 @@ pragma solidity 0.6.10;
...
+    struct ActionArgs {
+        /// @notice The type of action that is being performed on the system
+        ActionType actionType;
+        /// @notice The address of the account owner
+        address owner;
+        /// @notice The address which we move assets from or to (depending on the action type)
+        address sender;
+        /// @notice The asset that is to be transfered
+        address asset;
+        /// @notice The address of the vault that is to be modified (if any)
```
antoncoding, line 33: The `id` of the vault

contracts/libs/Actions.sol
```diff
@@ -8,5 +8,80 @@ pragma solidity 0.6.10;
...
+         */
+        uint256 index;
+        /// @notice Any other data that needs to be passed in for arbitrary function calls
+        bytes data;
+    }
+
+    struct DepositArgs {
+        /// @notice The address of the account owner
+        address owner;
+        /// @notice The address of the vault to which the asset will be added
```
antoncoding, line 49: the id of the vault

contracts/libs/Actions.sol
```diff
@@ -8,5 +8,80 @@ pragma solidity 0.6.10;
...
+    }
+
+    /**
+     * @notice Parses the passed in action argmuents to get the argmuents for a deposit action
+     * @param _args The general action arguments structure
+     * @return The arguments for a deposit action
+     */
+    function _parseDepositArgs(ActionArgs memory _args) internal returns (DepositArgs memory) {
+        require(
+            (uint256(_args.actionType) == uint256(ActionType.DepositLongOption)) ||
```
antoncoding, line 71: Why not just `_args.actionType == ActionType.DepositLongOption` like dydx does?

contracts/mocks/MockController.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma experimental ABIEncoderV2;
+
+import {Actions} from "../libs/Actions.sol";
+
+contract MockController {
```
antoncoding, line 6: We shouldn't call this `MockController` since it's just a contract for library testing, not mimicking any operation of the Controller.

contracts/mocks/MockController.sol
```diff
@@ -0,0 +1,16 @@
...
+    Actions.DepositArgs private depositArgs;
+
+    function testParseDespositAction(Actions.ActionArgs memory _args) external {
+        depositArgs = Actions._parseDepositArgs(_args);
```
antoncoding, line 10: Can't we just use `return Actions._parseDepositArgs` here and get rid of the `getDepositArgs`? 
This will make the test a little bit easier

test/Actions.test.ts
```diff
@@ -0,0 +1,121 @@
...
+        data: bytesArgs,
+      }
+
+      await expectRevert(
+        controller.testParseDespositAction(data),
+        'Actions: can only parse arguments for deposit actions',
+      )
+    })
+    it('should not be able to parse an invalid sender address', async () => {
+      const actionType = '3'
```
antoncoding, line 42: Given that we will have so many actions in this file in the future, will be better to define a Enum now.

test/Actions.test.ts
```diff
@@ -0,0 +1,121 @@
...
+        owner: owner,
+        sender: random,
+        asset: asset,
+        vaultId: vaultId,
+        amount: amount,
+        index: index,
+        data: bytesArgs,
+      }
+
+      await controller.testParseDespositAction(data)
```
antoncoding, line 81: Similar issue, calling this controller is confusing.

contracts/libs/Actions.sol
```diff
@@ -8,5 +8,80 @@ pragma solidity 0.6.10;
...
+        require(_args.owner != address(0), "Actions: cannot deposit to an invalid account");
+
+        DepositArgs memory args;
+        args.owner = _args.owner;
+        args.vaultId = _args.vaultId;
+        args.from = _args.sender;
+        args.asset = _args.asset;
+        args.index = _args.index;
+        args.amount = _args.amount;
+
```
antoncoding, line 84: Would [this syntax from dydx](https://github.com/dydxprotocol/solo/blob/3b6db45cf109fa2cf272843c4192a7d892ad30f4/contracts/protocol/lib/Actions.sol#L265) be more readable?
```
return DepositArgs({
            amount: args.amount,
            account: accounts[args.accountId],
            market: args.primaryMarketId,
            from: args.otherAddress
        });
```

contracts/mocks/MockController.sol
antoncoding: I suggest calling in `ActionTester`

contracts/mocks/MockController.sol
aparnakr: can't get return data while testing in truffle unless I emit an event. So either I should emit event or have this getter function. To test the state changes I decided to use the getter function. Event may mean that it gets emitted but doesn't change state

test/Actions.test.ts
aparnakr: hmm what do you mean?

contracts/libs/Actions.sol
aparnakr: hmm that changes how we have been doing return syntax so far. @haythem96 thoughts?

test/Actions.test.ts
antoncoding: Instead of using 
```javascript
const actionType = '3'
```
Have 
```javascript
// Defined globally
enum ActionType {
  OpenVault,
  MintShort,
  BurnShort,
  Deposit 
}

// In each test
const actionType = ActionType.Deposit
```

aparnakr: ohh you mean in the test file?

antoncoding:
One other thing, do you guys think we need another folder called `test-helper` in `contracts/`? We can move all the `__Test.sol` files there to make it clearer.

test/Actions.test.ts
```diff
@@ -0,0 +1,133 @@
...
+  }
+  const ZERO_ADDR = '0x0000000000000000000000000000000000000000'
+
+  before('Deployment', async () => {
+    actionTester = await ActionTester.new({from: owner})
+  })
+
+  describe('Parse Deposit Arguments', () => {
+    it('should not be able to parse a non Deposit action', async () => {
+      const actionType = ActionType.OpenVault
```
haythem96, line 30: Use `let` instead of `const` inside `it` blocks.

aparnakr:
> One other thing, do you guys think we need another folder called `test-helper` in `contracts/`? We can move all the `__Test.sol` files there to make it clearer.

we can maybe start with it being in mocks for now and then if it grows big move it?

test/Actions.test.ts
aparnakr: I think it is better practice to use const if we are not changing the value. The linter keeps changing it back when I change it to const

PR #33, opened 2020-07-23
WETH implementation

haythem96:
This PR add the needed code to wrap and unwrap ETH for users using [gnosis canonical-weth package](https://github.com/gnosis/canonical-weth).

In the case of transferring ETH into the system, the `Controller` will call the `deposit()` in WETH9 contract to wrap the ETH,  then treat as any other ERC20 and call `transferToPool()` passing the `address(this)` as sender.

In the case of transferring ETH out of the system, the `Controller` will call `transferToUser()` in the `MarginPool` passing `address(this)` as user arg, then call the `withdraw()` in `weth9` to unwrap the ETH and then transfer it out to user.

Will add more details in the spec.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,7 @@ jobs:
     working_directory: ~/gammaprotocol
     docker:
       - image: circleci/node:10.18.0
-      - image: trufflesuite/ganache-cli:v6.7.0
+      - image: trufflesuite/ganache-cli:v6.9.1
         command: ganache-cli --deterministic -e 300 -p 8545 -m 'candy maple cake sugar pudding cream honey rich smooth crumble sweet treat' --accounts 30 --allowUnlimitedContractSize
     
     steps:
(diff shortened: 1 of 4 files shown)


aparnakr:
Thanks for adding this @haythem96! So what module will call this?

haythem96:
> Thanks for adding this @haythem96! So what module will call this?

The `MarginPool` module will call this if the tranfer function call from the `Controller` have a `msg.value > 0` to wrap the sent ETH into WETH, then basically use the `withdraw` function when transferring out of the pool, to unwrap the WETH.

Will add the details into the spec

haythem96:
> Thanks for adding this @haythem96! So what module will call this?

Added that now in the description above.

antoncoding:
* Are we deploying our own WETH contract? Otherwise why do we it them in the `packages` folder and writing tests for it?
If we're just using the existing WETH contract, we could just put this in `mocks`.

antoncoding:
And I think we should just use existing WETH, so people already have WETH can also deposit in our system

haythem96:
> * Are we deploying our own WETH contract? Otherwise why do we it them in the `packages` folder and writing tests for it?
>   If we're just using the existing WETH contract, we could just put this in `mocks`, and don't need the tests

We are not deploying our own WETH, we will use an already deployed address. But even with that, users will send ETH and we will do wrapping internally.

re "If we're just using the existing WETH contract, we could just put this in mocks, and don't need the tests", I don't get that ?

PR #34, opened 2020-07-23
AddressBook module implementation

haythem96:
This PR add the `AddressBook` module functionalities and the needed upgrade contract for the proxy pattern using unstructured storage.


contracts/AddressBook.sol
```diff
@@ -3,10 +3,116 @@
...
+        OwnedUpgradeabilityProxy proxy = OwnedUpgradeabilityProxy(proxyAddress);
+        bytes memory params = abi.encodeWithSignature("initialize(address)", address(this));
 
+        if (proxyAddress == address(0)) {
+            proxy = new OwnedUpgradeabilityProxy();
+            proxy.upgradeToAndCall(_newAddress, params);
+            setAddress(_id, address(proxy));
+            emit ProxyCreated(_id, address(proxy));
+        } else {
+            proxy.upgradeToAndCall(_newAddress, params);
```
aparnakr, line 115: So this is setting a particular implementation and will only be called if the implementation is upgraded / a new contract not in the module is created?

contracts/AddressBook.sol
```diff
@@ -3,10 +3,119 @@
...
+    function getOracle() external view returns (address) {
+        return getAddress(ORACLE);
+    }
+
+    function setAddress(bytes32 _key, address _module) public onlyOwner {
+        modules[_key] = _module;
+    }
+
+    function setOtokenImpl(address _otokenImpl) external onlyOwner {
+        updateImplInternal(OTOKEN_IMPL, _otokenImpl);
```
antoncoding, line 70: We don't need proxy for `otokenImpl`. The `Otoken.sol` contract is not upgradable.

contracts/AddressBook.sol
```diff
@@ -3,10 +3,119 @@
...
+        OwnedUpgradeabilityProxy proxy = OwnedUpgradeabilityProxy(proxyAddress);
+        bytes memory params = abi.encodeWithSignature("initialize(address)", address(this));
 
+        if (proxyAddress == address(0)) {
+            proxy = new OwnedUpgradeabilityProxy();
+            proxy.upgradeToAndCall(_newAddress, params);
+            setAddress(_id, address(proxy));
+            emit ProxyCreated(_id, address(proxy));
+        } else {
+            proxy.upgradeToAndCall(_newAddress, params);
```
antoncoding, line 118: Refactor to

```javascript
address payable proxyAddress = address(uint160(getAddress(_id)));
bytes memory params = abi.encodeWithSignature("initialize(address)", address(this));
if (proxyAddress == address(0)) {
     OwnedUpgradeabilityProxy proxy = new OwnedUpgradeabilityProxy();
     proxy.upgradeToAndCall(_newAddress, params);
     setAddress(_id, address(proxy));
     emit ProxyCreated(_id, address(proxy));
} else {
       OwnedUpgradeabilityProxy proxy = OwnedUpgradeabilityProxy(proxyAddress);
       proxy.upgradeToAndCall(_newAddress, params);
 } 
```

haythem96: it looks like the written code is a refactor of ur comment ?!

antoncoding: in the `if` block, you declare a new proxy instance, 
so you should just put this line
`OwnedUpgradeabilityProxy proxy = OwnedUpgradeabilityProxy(proxyAddress);` in the else block?

contracts/AddressBook.sol
haythem96: yeah basically this function will get a module key and a deployed address, and will create an upgradeable proxy for it, if there is no V1 deployed, otherwise will upgrade V1 implementation to the new one (V2).

The `describe('Upgrade contract')` in the tests explain more the upgrading flow, and will share a doc soon about it.

test/addressBook.test.ts
```diff
@@ -0,0 +1,305 @@
...
+
+const MockERC20 = artifacts.require('MockERC20.sol')
+const OtokenFactory = artifacts.require('OtokenFactory.sol')
+const MarginPool = artifacts.require('MarginPool.sol')
+const UpgradeableContractV1 = artifacts.require('UpgradeableContractV1.sol')
+const UpgradeableContractV2 = artifacts.require('UpgradeableContractV2.sol')
+const OwnedUpgradeabilityProxy = artifacts.require('OwnedUpgradeabilityProxy.sol')
+const AddressBook = artifacts.require('AddressBook.sol')
+const Whitelist = artifacts.require('Whitelist.sol')
+// address(0)
```
aparnakr, line 21:
```suggestion
```

test/addressBook.test.ts
```diff
@@ -0,0 +1,305 @@
...
+      await expectRevert(
+        addressBook.setWhitelist(whitelist.address, {from: random}),
+        'Ownable: caller is not the owner',
+      )
+    })
+
+    it('should set whitelist address', async () => {
+      await addressBook.setWhitelist(whitelist.address, {from: owner})
+
+      assert.equal(await addressBook.getWhitelist(), whitelist.address, 'Whitelist address mismatch')
```
aparnakr, line 116: does it revert if you add a random address that is not a whitelistImplementation contract?

test/addressBook.test.ts
```diff
@@ -0,0 +1,305 @@
...
+const OtokenFactory = artifacts.require('OtokenFactory.sol')
+const MarginPool = artifacts.require('MarginPool.sol')
+const UpgradeableContractV1 = artifacts.require('UpgradeableContractV1.sol')
+const UpgradeableContractV2 = artifacts.require('UpgradeableContractV2.sol')
+const OwnedUpgradeabilityProxy = artifacts.require('OwnedUpgradeabilityProxy.sol')
+const AddressBook = artifacts.require('AddressBook.sol')
+const Whitelist = artifacts.require('Whitelist.sol')
+// address(0)
+
+contract('AddressBook', ([owner, otokenImplAdd, marginPoolAdd, random]) => {
```
aparnakr, line 23:
```suggestion
contract('AddressBook', ([owner, otokenImplAddress, marginPoolAddress, random]) => {
```

test/addressBook.test.ts
```diff
@@ -0,0 +1,305 @@
...
+  describe('Set margin pool', () => {
+    it('should revert adding pool address from non-owner address', async () => {
+      await expectRevert(addressBook.setMarginPool(marginPoolAdd, {from: random}), 'Ownable: caller is not the owner')
+    })
+
+    it('should set pool address', async () => {
+      await addressBook.setMarginPool(marginPoolAdd, {from: owner})
+
+      assert.equal(await addressBook.getMarginPool(), marginPoolAdd, 'Pool address mismatch')
+    })
```
aparnakr, line 129: why is this not testing with a newly deployed marginPool instance?

test/addressBook.test.ts
```diff
@@ -0,0 +1,305 @@
...
+    weth = await MockERC20.new('WETH', 'WETH')
+    // deploy AddressBook token
+    addressBook = await AddressBook.new()
+  })
+
+  describe('Set otoken implementation address', () => {
+    it('should revert adding otoken implementation address from non-owner address', async () => {
+      await expectRevert(addressBook.setOtokenImpl(otokenImplAdd, {from: random}), 'Ownable: caller is not the owner')
+    })
+
```
aparnakr, line 40: for oTokens, the owner is responsible for ensuring that a valid address is put in? The contract won't fail if I put in the a random addrres?

contracts/AddressBook.sol
```diff
@@ -3,10 +3,233 @@
...
+    }
+
+    /**
+     * @notice set WETH address
+     * @dev can only be called by addressbook owner
+     * @param _weth weth address
+     */
+    function setWeth(address _weth) external onlyOwner {
+        setAddress(WETH, _weth);
+    }
```
aparnakr, line 213: should this be here or in the whitelist module?

test/addressBook.test.ts
```diff
@@ -0,0 +1,305 @@
...
+      assert.equal(v2Controller.address, v2ImplementationAddress, 'Controller V2 implementation address mismatch')
+      assert.equal(v1Proxy.address, v2Proxy.address, 'Controller proxy address mismatch')
+
+      v2Controller = await UpgradeableContractV2.at(await addressBook.getController())
+
+      assert.equal(
+        (await v2Controller.CONTRACT_REVISION()).toString(),
+        '2',
+        'Controller implementation version mismatch',
```
aparnakr, line 301: So this is testing how you would upgrade the controller from v1 to v2? The address book itself is not upgradable right?

test/addressBook.test.ts
haythem96: why it should revert ? how does it know if that address is a whitelist implementation contract or not ?!

test/addressBook.test.ts
haythem96: MarginPool contract in this PR is not implemented yet, so I used an address, it behave the same as any address as the MarginPool is not upgradeable contract.

test/addressBook.test.ts
haythem96: yes, for all addresses, owner is responsible for making sure that `address X` is related to that contract/module.
The Owner in the beginning is the Opyn team through the Timelock contract, later will be a governance module.

contracts/AddressBook.sol
haythem96: It should be here as the `AddressBook` is the addresses provider for the system.

test/addressBook.test.ts
haythem96: yes this is testing upgrading from V1 to V2, the AddressBook is not upgradeable and does not need to be upgradeable in my opinion.

.solcover.js
```diff
@@ -14,6 +14,10 @@ module.exports = {
     'packages/oz/upgradeability/ContextUpgradeSafe.sol',
     'packages/oz/upgradeability/ERC20Initializable.sol',
     'packages/oz/upgradeability/Initializable.sol',
+    'packages/openzeppelin-upgradeability/OwnedUpgradeabilityProxy.sol',
+    'packages/openzeppelin-upgradeability/Proxy.sol',
+    'packages/openzeppelin-upgradeability/UpgradeabilityProxy.sol',
+    'packages/openzeppelin-upgradeability/VersionedInitializable.sol',
```
antoncoding, line 20: What's the differences between `oz/upgradeability` and `openzeppelin-upgradeability`? I think we can just use the same folder.

contracts/AddressBook.sol
```diff
@@ -3,10 +3,233 @@
...
+     * @notice set a specific address for a specific key
+     * @dev can only be called by addressbook owner
+     * @param _key key
+     * @param _address address
+     */
+    function setAddress(bytes32 _key, address _address) public onlyOwner {
+        addresses[_key] = _address;
+
+        emit AddressAdded(_key, _address);
+    }
```
antoncoding, line 132: We should put this after external function.

- Recommended order of functions in a contract: 
    - constructor
    - fallback
    - external
    - public
    - internal
    - private

contracts/AddressBook.sol
antoncoding: @haythem96 thoughts?

contracts/tests/UpgradeableContractV1.sol
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    /**
+     * @dev this function is invoked by the proxy contract when this contract is added to the
+     * AddressBook.
+     * @param _addressBook the address of the AddressBook
+     **/
+    function initialize(address _addressBook) public initializer {
+        addressBook = _addressBook;
+    }
```
antoncoding, line 33: Just to make sure, does this mean all the upgradable contract in our system needs to have the same `initialize` interface, with only one argument (`_addressbook`) ?

test/addressBook.test.ts
```diff
@@ -0,0 +1,304 @@
...
+  })
+
+  describe('Set controller', () => {
+    let controller: UpgradeableContractV1Instance
+
+    before(async () => {
+      controller = await UpgradeableContractV1.new()
+    })
+
+    it('should revert adding otoken implementation address from non-owner address', async () => {
```
antoncoding, line 54:
```suggestion
    it('should revert adding controller from non-owner address', async () => {
```

test/addressBook.test.ts
```diff
@@ -0,0 +1,304 @@
...
+      await addressBook.setController(controller.address, {from: owner})
+
+      const proxy: OwnedUpgradeabilityProxyInstance = await OwnedUpgradeabilityProxy.at(
+        await addressBook.getController(),
+      )
+      const implementationAddress = await proxy.implementation()
+
+      assert.equal(controller.address, implementationAddress, 'Controller implementation address mismatch')
+
+      controller = await UpgradeableContractV1.at(await addressBook.getController())
```
antoncoding, line 71:
```suggestion
      controller = await UpgradeableContractV1.at(proxy.address)
```

antoncoding, line 71:
```suggestion
```
I don't think you need this line?
controller is already `UpgradeableContractV1`, and getController() return `controller.address`

test/addressBook.test.ts
```diff
@@ -0,0 +1,304 @@
...
+      await addressBook.setMarginCalculator(marginCalculator.address, {from: owner})
+
+      const proxy: OwnedUpgradeabilityProxyInstance = await OwnedUpgradeabilityProxy.at(
+        await addressBook.getMarginCalculator(),
+      )
+      const implementationAddress = await proxy.implementation()
+
+      assert.equal(marginCalculator.address, implementationAddress, 'Margin calculator implementation address mismatch')
+
+      marginCalculator = await UpgradeableContractV1.at(await addressBook.getMarginCalculator())
```
antoncoding, line 155: Same issue here and for `liquidationManager `

contracts/tests/UpgradeableContractV1.sol
haythem96: yes.

test/addressBook.test.ts
haythem96: yeah but `await UpgradeableContractV1.new()` and `await addressBook.getController()` are two different addresses, the `getController()` return the proxy address.

contracts/AddressBook.sol
haythem96: the line u added in the `else` block is the L109 in the current implementation, for me the current implementation looks like a refactor for what u wrote.
Also is this so important ? does it make an issue in term of execution ?

haythem96:
> Re: folder structure. It would be a pain if we need to change packages paths in the future, let's just decide this now in this PR.
> 
> * What's the difference between `oz/upgradeability` and `openzeppelin-upgradeability`?

move the contracts into `oz/upgradeability`

test/addressBook.test.ts
antoncoding: Okay got it now ;)

contracts/AddressBook.sol
antoncoding: Yeah I'm suggesting putting that line in the else block.

antoncoding: I just feel like you don't need L109 outside of the if-else condition, because if (proxyAddress==0), you'll just waste the gas that executes L109 right?

haythem96: okay I changed that.

PR #35, opened 2020-07-24
Pool module: single address transfer

haythem96:
This PR add the needed code for transferring funds  in and out the pool.
Transferring multiple assets + multiple addresses will be added in a separate PR.

### DIFF
--- a/contracts/MarginPool.sol
+++ b/contracts/MarginPool.sol
@@ -3,10 +3,65 @@
  */
 pragma solidity 0.6.10;
 
+import {ERC20Interface} from "./interfaces/ERC20Interface.sol";
+import {AddressBookInterface} from "./interfaces/AddressBookInterface.sol";
+import {SafeMath} from "./packages/oz/SafeMath.sol";
+
 /**
  *
  */
-// solhint-disable-next-line no-empty-blocks
 contract MarginPool {
+    using SafeMath for uint256;
 
+    uint256 public constant BASE_UNIT = 1e18;
+
+    /// @notice AddressBook module
+    address public addressBook;
+
+    /**
+     * @notice contructor
+     * @param _addressBook adressbook module
+     */
+    constructor(address _addressBook) public {
+        require(_addressBook != address(0), "Invalid address book");
+
+        addressBook = _addressBook;
+    }
+
+    /**
+     * @notice check if the sender is the Controller module
+     */
+    modifier onlyController() {
+        require(
+            msg.sender == AddressBookInterface(addressBook).getController(),
+            "MarginPool: Sender is not Controller"
+        );
+
+        _;
+    }
+
+    function transferToPool(
+        address _asset,
+        address _user,
+        uint256 _amount
+    ) external payable onlyController returns (bool success) {
+        require(_amount > 0, "MarginPool: invalid amount");
+
+        // get asset decimals
+        uint8 assetDecimal = ERC20Interface(_asset).decimals();
+        // scale amount
+        uint256 val = _calcTransferAmount(_amount, assetDecimal);
+        // transfer val from _user to pool
+        ERC20Interface(_asset).transferFrom(_user, address(this), val);
+    }
+
+    /**
+     * @notice Scale _amt
+     * @param _amt amount of tokens
+     * @param _decimals token decimals
+     * @return scaled amount
+     */
+    function _calcTransferAmount(uint256 _amt, uint256 _decimals) internal returns (uint256) {
+        return _amt.div(BASE_UNIT.sub(_decimals));
+    }
 }
--- a/contracts/MarginPool.sol
+++ b/contracts/MarginPool.sol
@@ -22,9 +22,6 @@ contract MarginPool {
     /// @notice AddressBook module
     address public addressBook;
 
-    /// @notice WETH token
-    WETH9 public WETH;
-
     /**
      * @notice contructor
      * @param _addressBook adressbook module
@@ -33,8 +30,6 @@ contract MarginPool {
         require(_addressBook != address(0), "Invalid address book");
 
         addressBook = _addressBook;
-
-        WETH = WETH9(payable(AddressBookInterface(addressBook).getWeth()));
     }
 
     /**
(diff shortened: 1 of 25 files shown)


contracts/interfaces/AddressBookInterface.sol
```diff
@@ -4,6 +4,8 @@ pragma solidity 0.6.10;
 interface AddressBookInterface {
     /* Getters */
 
+    function getWethToken() external view returns (address);
+
```
aparnakr, line 8: should this be here or in whitelist module?

contracts/MarginPool.sol
```diff
@@ -3,10 +3,135 @@
...
+        }
+    }
 
+    /**
+     * @notice Scale _amt
+     * @param _amt amount of tokens
+     * @param _decimals token decimals
+     * @return scaled amount
+     */
+    function _calcTransferAmount(uint256 _amt, uint256 _decimals) internal pure returns (uint256) {
```
haythem96, line 134: @aparnakr @antoncoding this function should be here, right ?

antoncoding: No I don't think we need this, for transferToUser, we can calculate the amount in controller and call `Pool.transferToUser()`, for transferToPool, we can call it with the amount value in the arguments (un-scaled)

contracts/MarginPool.sol
```diff
@@ -3,10 +3,135 @@
...
+        // get asset decimals
+        uint8 assetDecimal = ERC20Interface(_asset).decimals();
+        // scale amount
+        uint256 val = _calcTransferAmount(_amount, assetDecimal);
+
+        bool success;
+
+        /// check if asset is WETH or not, if WETH transfer the val from Controller address
+        if (_asset == address(WETH)) {
+            // tranfer WETH from controller to pool
```
aparnakr, line 83: where does wrapping of the ETH happen? In what contract?

contracts/MarginPool.sol
```diff
@@ -3,10 +3,135 @@
...
+     * @param _asset address of asset to transfer
+     * @param _user address of user to transfer assets from
+     * @param _amount amount of token to transfer from _user, scaled to 1e18 of precision
+     * @return true if successful transfer
+     */
+    function transferToPool(
+        address _asset,
+        address _user,
+        uint256 _amount
+    ) external onlyController returns (bool) {
```
aparnakr, line 71:
```suggestion
    ) external onlyController {
```

contracts/MarginPool.sol
```diff
@@ -3,10 +3,135 @@
...
+        /// check if asset is WETH or not, if WETH transfer the val from Controller address
+        if (_asset == address(WETH)) {
+            // tranfer WETH from controller to pool
+            success = ERC20Interface(_asset).transferFrom(msg.sender, address(this), val);
+        } else {
+            // transfer val from _user to pool
+            success = ERC20Interface(_asset).transferFrom(_user, address(this), val);
+        }
+
+        return success;
```
aparnakr, line 90:
```suggestion
        return success;
```

contracts/MarginPool.sol
```diff
@@ -3,10 +3,135 @@
...
+
+        bool success;
+
+        /// check if asset is WETH or not, if WETH transfer the val from Controller address
+        if (_asset == address(WETH)) {
+            // tranfer WETH from controller to pool
+            success = ERC20Interface(_asset).transferFrom(msg.sender, address(this), val);
+        } else {
+            // transfer val from _user to pool
+            success = ERC20Interface(_asset).transferFrom(_user, address(this), val);
```
aparnakr, line 87: can this fail if the transfer doesn't happen?

contracts/MarginPool.sol
```diff
@@ -3,10 +3,135 @@
...
+     * @param _asset address of asset to transfer
+     * @param _user address of user to transfer assets to
+     * @param _amount amount of token to transfer to _user, scaled to 1e18 of precision
+     * @return true if successful transfer
+     */
+    function transferToUser(
+        address _asset,
+        address payable _user,
+        uint256 _amount
+    ) external onlyController returns (bool) {
```
aparnakr, line 106:
```suggestion
    ) external onlyController {
```

contracts/MarginPool.sol
```diff
@@ -3,10 +3,135 @@
...
+        if (_asset == address(WETH)) {
+            // unwrap WETH
+            WETH.withdraw(val);
+            // transfer ETH to user
+            _user.transfer(val);
+
+            return true;
+        } else {
+            // transfer asset val from Pool to _user
+            return ERC20Interface(_asset).transfer(_user, val);
```
aparnakr, line 124: fail if this doesn't happen

contracts/MarginPool.sol
haythem96: cool, I will update this. cc @aleone

contracts/MarginPool.sol
haythem96: in the controller level, somewhere in the operate function, should be on of the first steps.

contracts/MarginPool.sol
haythem96: yeah there are require statement in the WETH contract, also we can add require statement in the controller level to check that pool always return true. IMO make sense to add those return checks in the controller level.

contracts/MarginPool.sol
haythem96: yeah

contracts/MarginPool.sol
haythem96: so @antoncoding the amount in the args params will be by default scaled, in the case of `transferToUser()` ?

antoncoding: yeah, in both cases, you don't have to deal with scaling.

contracts/MarginPool.sol
antoncoding: Right now the flow looks like following:

* Transfer to pool: `Controller` call WETH.deposit, and `Pool` pull tokens from Controller
* Transfer to user: `Pool` calls WETH.withdraw, and transfer ETH to user

I don't see a clear reason why we need to use `WETH` instead of transferring ETH in this case, because both Controller and Pool are having checks for ETH asset, so we are not simplifying anything IMO. 

I imagined using WETH would let us remove ETH check in one of the contracts, right now it's just the same as using ETH natively

haythem96: We used WETH so we don't use address(0) for ETH, as there are so many checks that use `address(0)` in the MarginAccount...etc

contracts/MarginPool.sol
aparnakr: why not just fail here instead of another check at the controller level?

aparnakr: that way we know the controller will not have to worry about another check there right ?

haythem96: I don't know, thought from consistency perspective, all other module return result and Controller make the check.

antoncoding: IMO not necessary, making the check from Controller also make sense.

PR #38, opened 2020-07-25
Integration test - open markets

### LINKED ISSUE #28: Functional tests for oToken + oTokenFactory

antoncoding:
## Open Markets tests
closes #28
### High level description of the tests
- [x] After admin whitelist product, anyone can create product on arbitrary strikePrice, expiry.
- [x] The oTokens deployed by the factory pass all the property test.
- [x] After the Address book update the otokens, factory should deploy otokens with the new template.

### Bug fix
* Fix the wrong constructor usage in `Otoken.sol`


test/OtokenFactoryIntegration.test.ts
```diff
@@ -0,0 +1,188 @@
...
+  let usdc: MockERC20Instance
+  let dai: MockERC20Instance
+  let randomERC20: MockERC20Instance
+
+  const ethAddress = ZERO_ADDR
+  const strikePrice = new BigNumber(200).times(new BigNumber(10).exponentiatedBy(18))
+  const isPut = true
+  const expiry = 1753776000 // 07/29/2025 @ 8:00am (UTC)
+
+  before('Deploy addreeBook, otoken logic, whitelist, Factory contract', async () => {
```
aparnakr, line 35:
```suggestion
  before('Deploy addressBook, otoken logic, whitelist, Factory contract', async () => {
```

test/utils.ts
```diff
@@ -0,0 +1,28 @@
...
+
+const OTokenFactory = artifacts.require('OtokenFactory.sol')
+const MockAddressBook = artifacts.require('MockAddressBook.sol')
+const Whitelist = artifacts.require('Whitelist.sol')
+const Otoken = artifacts.require('Otoken.sol')
+
+export async function setupContracts(deployer: string) {
+  // Todoo: Use Real addressBook instance when available
+
+  const addressBook: MockAddressBookInstance = await MockAddressBook.new({from: deployer})
```
aparnakr, line 16: can we not use the mock instance and use the real one?

aparnakr, line 16: can we use the real instance once it is done? this is a little confusing to keep track of what is mock and what is not

antoncoding:
> Let's maybe wait on this PR until addressbook is done / we figure out a testing practice? I'm getting a little confused by where we are testing with mocks and where it is the real code and that can potentially lead to some bigger issues down the line
@aparnakr Yeah we will use real instances in functional test, I open this now hopefully to make it clear on "Which tests are functional tests", and shouldn't be in scope of one module development.

antoncoding:
### Mock vs Real instacnes
For now, I still need to use a `TestController` contract to test a `mintOtoken` and `burnOtoken` flow from the Controller, (I basically just need a contract that calls `mintOtoken` and `burnOtokens`, and they have to be contracts instead of EOAs because the Addressbook module will deploy a proxy and use that as the "Controller address") 

I do think these test helper contracts are essential, otherwise I won't be able to test the the whole otoken flow until we have our controller ready, and we don't have functions that directly call mint and burn on Otoken contracts, so the test would ends up calling `operate` function with Mint and Burn operations, which will involve setting up oracle just to test the mint / burn permission.

But if everyone think we should enforce using real instances in all integration test for better consistency, I can remove this part for now and add that later when we got our Controller ready ;)

contracts/tests/TestController.sol
```diff
@@ -0,0 +1,47 @@
...
+pragma solidity 0.6.10;
+
+// import "../packages/oz/upgradeability/VersionedInitializable.sol";
+import "../interfaces/OtokenInterface.sol";
+
+/**
+ * @author Opyn Team
+ * @notice Upgradeable Controller testing contract
+ */
+contract TestController {
```
aparnakr, line 13: hmm after looking at how a bunch of other teams do it, maybe we want to call this also a MockController contract? A TestController Id imagine inherits the controller and tests it?

test/Otoken.test.ts
```diff
@@ -50,125 +52,155 @@ contract('Otoken', ([deployer, controller, user1, user2, random]) => {
...
+        otoken.init(addressBookAddr, usdc.address, ETH_ADDR, ETH_ADDR, strikePrice, expiry, false),
         'Contract instance has already been initialized.',
       )
     })
 
     it('should set the right name for calls', async () => {
-      const call = await Otoken.new(mockAddressBookAddr)
-      await call.init(ETH_ADDR, usdc.address, usdc.address, strikePrice, expiry, false, {from: deployer})
+      const call = await Otoken.new(addressBookAddr)
+      await call.init(addressBookAddr, ETH_ADDR, usdc.address, usdc.address, strikePrice, expiry, false, {
```
aparnakr, line 69:
```suggestion
      const callOption = await Otoken.new(addressBookAddr)
      await callOption.init(addressBookAddr, ETH_ADDR, usdc.address, usdc.address, strikePrice, expiry, false, {
```

test/Otoken.test.ts
```diff
@@ -50,125 +52,155 @@ contract('Otoken', ([deployer, controller, user1, user2, random]) => {
...
       assert.equal(await call.symbol(), `oETHUSDC-25SEP20-200C`)
     })
 
     it('should set the right name for non-eth options', async () => {
       const weth = await MockERC20.new('WETH', 'WETH')
-      const put = await Otoken.new(mockAddressBookAddr)
-      await put.init(weth.address, usdc.address, usdc.address, strikePrice, expiry, isPut, {from: deployer})
+      const put = await Otoken.new(addressBookAddr)
+      await put.init(addressBookAddr, weth.address, usdc.address, usdc.address, strikePrice, expiry, isPut, {
+        from: deployer,
```
aparnakr, line 80:
```suggestion
      const putOption = await Otoken.new(addressBookAddr)
      await putOption.init(addressBookAddr, weth.address, usdc.address, usdc.address, strikePrice, expiry, isPut, {
        from: deployer,
```

test/OtokenFactory.test.ts
```diff
@@ -182,39 +183,38 @@ contract('OTokenFactory', ([user1, user2]) => {
...
         {from: user2},
       )
       expectEvent(txResponse, 'OtokenCreated', {
         creator: user2,
-        underlying: ethAddress,
-        strike: usdc.address,
-        collateral: usdc.address,
+        underlying: usdc.address,
+        strike: ethAddress,
+        collateral: ethAddress,
```
aparnakr, line 217: why change this?

test/marginPool.test.ts
```diff
@@ -37,7 +37,7 @@ contract('MarginPool', ([controllerAddress, user1, random]) => {
     // set Controller module address
     await addressBook.setController(controllerAddress)
     // set WETH address
-    await addressBook.setWeth(weth.address)
+    // await addressBook.setWeth(weth.address)
```
aparnakr, line 40:
```suggestion
```

aparnakr (approved):
LGTM!

PR #41, opened 2020-07-25
Actions Module: Open Vault Args

aparnakr:
This PR adds the following:
1. OpenVaultArgs struct
2. function to parse the openVault arguments

### DIFF
--- a/contracts/libs/Actions.sol
+++ b/contracts/libs/Actions.sol
@@ -68,7 +68,7 @@ library Actions {
     }
 
     /**
-     * @notice Parses the passed in action argmuents to get the argmuents for a open vault action
+     * @notice Parses the passed in action argmuents to get the argmuents for an open vault action
      * @param _args The general action arguments structure
      * @return The arguments for a open vault action
      */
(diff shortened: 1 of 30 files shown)


contracts/libs/Actions.sol
```diff
@@ -62,6 +67,18 @@ library Actions {
         uint256 amount;
     }
 
+    /**
+     * @notice Parses the passed in action argmuents to get the argmuents for a open vault action
```
antoncoding, line 71:
```suggestion
     * @notice Parses the passed in action argmuents to get the argmuents for an open vault action
```

contracts/packages/oz/Strings.sol
```diff
@@ -26,9 +26,9 @@ library Strings {
         uint256 index = digits - 1;
         temp = value;
         while (temp != 0) {
-            buffer[index--] = byte(uint8(48 + temp % 10));
+            buffer[index--] = bytes1(uint8(48 + (temp % 10)));
```
antoncoding, line 29: Why do we need to change this?

aparnakr: idk, linter added it lol

PR #43, opened 2020-07-26
Actions Module: Mint and Burn Args

aparnakr:
This PR adds the following:

MintArgs struct
BurnArgs struct
function to parse the mint arguments
function to parse the burn arguments


test/Actions.test.ts
```diff
@@ -130,4 +130,210 @@ contract('Actions', ([owner, random]) => {
...
+        owner: owner,
+        sender: owner,
+        asset: asset,
+        vaultId: vaultId,
+        amount: amount,
+        index: index,
+        data: bytesArgs,
+      }
+
+      await expectRevert(actionTester.testParseMintAction(data), 'Actions: can only parse arguments for mint actions')
```
antoncoding, line 154:
```suggestion
    it('should not be able to parse a non Mint action', async () => {
      const data = {
        actionType: ActionType.OpenVault,
        owner: ZERO_ADDR,
        sender: owner,
        asset: ZERO_ADDR,
        vaultId: vaultId,
        amount: '10',
        index: '0',
        data: ZERO_ADDR,
      }

      await expectRevert(actionTester.testParseMintAction(data), 'Actions: can only parse arguments for mint actions')
```

antoncoding: If we're not using those `const` variables for verification, then we can just declare the testing data in the `data` object, which save us some lines.

test/Actions.test.ts
```diff
@@ -130,4 +130,210 @@ contract('Actions', ([owner, random]) => {
...
+        asset: asset,
+        vaultId: vaultId,
+        amount: amount,
+        index: index,
+        data: bytesArgs,
+      }
+
+      await expectRevert(actionTester.testParseMintAction(data), 'Actions: cannot mint from an invalid account')
+    })
+    it('should be able to parse arguments for a mint long action', async () => {
```
antoncoding, line 177:
```suggestion
    it('should be able to parse arguments for a mint short action', async () => {
```

test/Actions.test.ts
```diff
@@ -130,4 +130,210 @@ contract('Actions', ([owner, random]) => {
...
+
+      const burnArgs = await actionTester.getBurnArgs()
+      assert.equal(burnArgs.owner, owner)
+      assert.equal(burnArgs.amount, new BN(amount))
+      assert.equal(burnArgs.otoken, asset)
+      assert.equal(burnArgs.from, random)
+      assert.equal(burnArgs.vaultId, new BN(vaultId))
+      assert.equal(burnArgs.index, new BN(index))
+    })
+    it('should be able to parse arguments for a burn collateral action', async () => {
```
antoncoding, line 309:
```suggestion
    it('should be able to parse arguments for a burn short action', async () => {
```

test/Actions.test.ts
```diff
@@ -130,4 +130,210 @@ contract('Actions', ([owner, random]) => {
...
+
+      const mintArgs = await actionTester.getMintArgs()
+      assert.equal(mintArgs.owner, owner)
+      assert.equal(mintArgs.amount, new BN(amount))
+      assert.equal(mintArgs.otoken, asset)
+      assert.equal(mintArgs.to, random)
+      assert.equal(mintArgs.vaultId, new BN(vaultId))
+      assert.equal(mintArgs.index, new BN(index))
+    })
+    it('should be able to parse arguments for a mint collateral action', async () => {
```
antoncoding, line 206: It shouldn't be a `mint collateral` test right? Not sure what this test is for, it seems similar to the previous one.

aparnakr, line 199:
```suggestion
    it('should be able to parse arguments for a mint short action', async () => {
```

aparnakr:
```suggestion
    it('should be able to parse arguments for a mint action', async () => {
```

test/Actions.test.ts
```diff
@@ -130,4 +130,203 @@ contract('Actions', ([owner, random]) => {
...
+        asset: asset,
+        vaultId: vaultId,
+        amount: amount,
+        index: index,
+        data: bytesArgs,
+      }
+
+      await expectRevert(actionTester.testParseBurnAction(data), 'Actions: cannot burn from an invalid account')
+    })
+    it('should be able to parse arguments for a burn long action', async () => {
```
aparnakr, line 273:
```suggestion
    it('should be able to parse arguments for a burn short action', async () => {
```

contracts/libs/Actions.sol
```diff
@@ -62,6 +98,46 @@ library Actions {
...
+     */
+    function _parseMintArgs(ActionArgs memory _args) internal returns (MintArgs memory) {
+        require(_args.actionType == ActionType.MintShortOption, "Actions: can only parse arguments for mint actions");
+        require(_args.owner != address(0), "Actions: cannot mint from an invalid account");
+
+        return
+            MintArgs({
+                owner: _args.owner,
+                vaultId: _args.vaultId,
+                to: _args.sender,
```
haythem96, line 114: what does `to: _args.sender` refer to ? so in this case the minted short option will be stored in the `vault.vaultId` and the tokens will be minted to the `sender` and not the `owner` right ? the sender can be operator also.

What happen in the scenario where an operator mint token for a vault, and he get the tokens, and then the vault owner want to do some operation with the short option amount, but he doesn't hold them, but still the short options are stored in his vault ?

aparnakr: yeah, that's why its important for the end user to be careful about who they authorize the vault to. The checks on sender is vault operator should all happen in the controller.

aparnakr: tbh the vault operator can also be the person who deposited collateral. So it works both ways

aparnakr: maybe initially we can whitelist those addresses that can be vault operators? @antoncoding

haythem96: yeah my point is I think minted option should always be minted to the vault owner. Operator will just operate vault, but will not get ownership of minted option.

antoncoding: Either way it's not a concern here, all the scenario you guys discuss will be handle in the Controller (check MintArg.to == owner or whatever), here we're just parsing the arguments from general Arg type to MintArgs.

aparnakr: @haythem96 any final thoughts or can I resolve this?

haythem96: @antoncoding can u elaborate more ? my concern is regarding minting short options to operator address, will that be possible or all minted short will be sent to owner ?

antoncoding: For now, in this library, we only convert general Args struct into `MintArgs`(parseMintArgs function). it will only does basic checks around invalid address, and the real access control logic should be done in the controller. 
```javascript
// Controller.sol
function mintShortOtokens(MintArgs args) {
    require(msg.sender == args.owner || isOperator[args.owner][msg.sender], "Tx sender must be owner or operator.")
    require(args.to == owner || args.to == msg.sender, "Cannot mint to arbitrary accounts") // if needed
}
```

I think those logics belong to the process of "minting tokens", instead of `parsing arguments`, so I would want to put these in the `mintShortOtokens` instead of `parseMintArgs`

haythem96: yeah, I'm not talking about the checks, so in those two requires, the operator can can call mint action, and get the minted options in his balance, what's ur thought re that ?!

haythem96: would love ur thoughts @zkoticha @aleone .

antoncoding: @haythem96 those 2 requirements are in Controller, so we should still be able to merge this now.

And if we want to empower a wrapper contract, I think it's essential to let the wrapper decide where to mint tokens to. If the wrapper can only mint token to the owner, it won't be able to play something like "flashloan", because after it mints the token to the owner's account, it can't move those tokens until the owner approve again.

contracts/libs/Actions.sol
```diff
@@ -44,6 +44,42 @@ library Actions {
...
+        address otoken;
+        /**
+         * @notice Each vault can hold multiple short / long / collateral assets. In this version, we are restricting the scope to only 1 of each.
+         * In future versions this would be the index of the short / long / collateral asset that needs to be modified.
+         */
+        uint256 index;
+        /// @notice The amount of otokens that is to be burned
+        uint256 amount;
+    }
+
```
antoncoding, line 82: Change `///@notice` doc string to normal comment `//`

PR #44, opened 2020-07-26
Actions Module: SettleVault Args

aparnakr:
This PR adds the following:

SettleVaultArgs struct
function to parse the settleVault arguments

### DIFF
--- a/contracts/libs/Actions.sol
+++ b/contracts/libs/Actions.sol
@@ -62,6 +62,15 @@ library Actions {
         uint256 amount;
     }
 
+    struct SettleVaultArgs {
+        /// @notice The address of the account owner
+        address owner;
+        /// @notice The index of the vault to which is to be settled
+        uint256 vaultId;
+        /// @notice The address to which we transfer the remaining collateral
+        address to;
+    }
+
     /**
      * @notice Parses the passed in action argmuents to get the argmuents for a deposit action
      * @param _args The general action arguments structure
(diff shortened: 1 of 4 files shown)


test/Actions.test.ts
```diff
@@ -130,4 +130,81 @@ contract('Actions', ([owner, random]) => {
...
+        index: index,
+        data: bytesArgs,
+      }
+
+      await expectRevert(
+        actionTester.testParseSettleVaultAction(data),
+        'Actions: can only parse arguments for settle vault actions',
+      )
+    })
+    it('should not be able to parse an invalid sender address', async () => {
```
antoncoding, line 159: > 
```suggestion
    it('should not be able to parse an invalid owner address', async () => {
```

contracts/libs/Actions.sol
```diff
@@ -84,4 +93,19 @@ library Actions {
                 amount: _args.amount
             });
     }
+
+    /**
+     * @notice Parses the passed in action argmuents to get the argmuents for a settle vault action
+     * @param _args The general action arguments structure
+     * @return The arguments for a settle vault action
+     */
+    function _parseSettleVaultArgs(ActionArgs memory _args) internal returns (SettleVaultArgs memory) {
```
haythem96, line 102: should this check if the sender is the vault owner or operator ? or do we do this check somewhere else ?

antoncoding: I think somewhere else because the library won't have access to the operator mapping right?

aparnakr: no that check is in the controller.

PR #46, opened 2020-07-28
[WIP] TimeLock module

haythem96:


haythem96:
> Is this just Compound's timelock package or are we modifying it?

It is Compound code, but we will need to modify it depend on the emergency shutdown implementation.

antoncoding:
Close this PR for now, will reopen when we need to add this module

PR #47, opened 2020-07-28
Refactor updateImpl()

haythem96:


contracts/AddressBook.sol
```diff
@@ -220,7 +220,6 @@ contract AddressBook is Ownable {
     function updateImpl(bytes32 _id, address _newAddress) public onlyOwner {
         address payable proxyAddress = address(uint160(getAddress(_id)));
 
-        OwnedUpgradeabilityProxy proxy = OwnedUpgradeabilityProxy(proxyAddress);
```
aparnakr, line 223: we don't need this anywhere?

haythem96: I refactored this function in the AddressBook PR #34  as Anton suggest but forgot to remove this line, we don't need it anywhere.

PR #48, opened 2020-07-28
FixedPointInt256 library

haythem96:
Int256 fixed point library.

### DIFF
--- a/contracts/libs/FixedPointInt256.sol
+++ b/contracts/libs/FixedPointInt256.sol
@@ -46,11 +46,11 @@ library FixedPointInt256 {
     }
 
     function mul(int256 a, int256 b) internal pure returns (int256) {
-        return (a.mul(b)).div(SCALING_FACTOR);
+        return add(mul(a, b), SCALING_FACTOR / 2) / SCALING_FACTOR;
     }
 
     function div(int256 a, int256 b) internal pure returns (int256) {
-        return (a.div(b)).mul(SCALING_FACTOR);
+        return add(mul(a, SCALING_FACTOR), b / 2) / b;
     }
 
     function min(int256 a, int256 b) internal pure returns (int256) {
--- a/contracts/libs/FixedPointInt256.sol
+++ b/contracts/libs/FixedPointInt256.sol
@@ -98,4 +98,54 @@ library FixedPointInt256 {
     function max(int256 a, int256 b) internal pure returns (int256) {
         return a > b ? a : b;
     }
+
+    /**
+     * @notice Whether `a` is equal to `b`.
+     * @param a a signed integer
+     * @param b a signed integer
+     * @return True if equal, or False.
+     */
+    function isEqual(int256 a, int256 b) internal pure returns (bool) {
+        return a == b;
+    }
+
+    /**
+     * @notice Whether `a` is greater than `b`.
+     * @param a a signed integer
+     * @param b a signed integer
+     * @return True if `a > b`, or False.
+     */
+    function isGreaterThan(int256 a, int256 b) internal pure returns (bool) {
+        return a > b;
+    }
+
+    /**
+     * @notice Whether `a` is greater than or equal to `b`.
+     * @param a a signed integer
+     * @param b a signed integer
+     * @return True if `a >= b`, or False.
+     */
+    function isGreaterThanOrEqual(int256 a, int256 b) internal pure returns (bool) {
+        return a >= b;
+    }
+
+    /**
+     * @notice Whether `a` is less than `b`.
+     * @param a a signed integer
+     * @param b a signed integer
+     * @return True if `a < b`, or False.
+     */
+    function isLessThan(int256 a, int256 b) internal pure returns (bool) {
+        return a < b;
+    }
+
+    /**
+     * @notice Whether `a` is less than or equal to `b`.
+     * @param a a signed integer
+     * @param b a signed integer
+     * @return True if `a <= b`, or False.
+     */
+    function isLessThanOrEqual(int256 a, int256 b) internal pure returns (bool) {
+        return a <= b;
+    }
 }
--- a/contracts/libs/FixedPointInt256.sol
+++ b/contracts/libs/FixedPointInt256.sol
@@ -1,0 +1,22 @@
+/**
+ * SPDX-License-Identifier: UNLICENSED
+ */
+pragma solidity 0.6.10;
+
+/**
+ *
+ */
+library FixedPointInt256 {
+    uint256 private constant SCALING_FACTOR = 1e18;
+
+    /**
+     * @notice convert an unsigned integer to signed integer
+     * @param a uint to convert into a signed integer.
+     * @return the converted signed integer.
+     */
+    function fromUint(uint256 a) internal pure returns (int256) {
+        require(a < uint256(-1), "FixedPointInt256: can't cast - out of int range");
+
+        return int256(a);
+    }
+}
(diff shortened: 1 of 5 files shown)


contracts/libs/FixedPointInt256.sol
```diff
@@ -0,0 +1,63 @@
...
+
+    int256 private constant SCALING_FACTOR = 1e18;
+
+    /**
+     * @notice convert an unsigned integer to signed integer
+     * @param a uint to convert into a signed integer.
+     * @return the converted signed integer.
+     */
+    function fromUint(uint256 a) internal pure returns (int256) {
+        require(a < uint256(-1), "FixedPointInt256: out of int range");
```
aparnakr, line 22: also a max value restriction right?

contracts/libs/FixedPointInt256.sol
```diff
@@ -0,0 +1,63 @@
...
+
+        return int256(a);
+    }
+
+    /**
+     * @notice convert a signed integer to unsigned integer
+     * @param a int to convert into an unsigned integer.
+     * @return the converted unsigned integer.
+     */
+    function fromInt(int256 a) internal pure returns (int256) {
```
aparnakr, line 32:
```suggestion
    function fromInt(int256 a) internal pure returns (uint256) {
```

contracts/libs/FixedPointInt256.sol
```diff
@@ -0,0 +1,63 @@
...
+    function sub(int256 a, int256 b) internal pure returns (int256) {
+        return a.sub(b);
+    }
+
+    function mul(int256 a, int256 b) internal pure returns (int256) {
+        return (a.mul(b)).div(SCALING_FACTOR);
+    }
+
+    function div(int256 a, int256 b) internal pure returns (int256) {
+        return (a.div(b)).mul(SCALING_FACTOR);
```
aparnakr, line 53: maybe you want to round these? See maker's math library for deets

haythem96: cool

contracts/libs/FixedPointInt256.sol
haythem96: yeah.

contracts/libs/FixedPointInt256.sol
```diff
@@ -0,0 +1,101 @@
...
+    using SignedSafeMath for int256;
+
+    int256 private constant SCALING_FACTOR = 1e18;
+
+    /**
+     * @notice convert an unsigned integer to signed integer
+     * @param a uint to convert into a signed integer.
+     * @return the converted signed integer.
+     */
+    function fromUint(uint256 a) internal pure returns (int256) {
```
aparnakr, line 21: can we name it uintToInt?

contracts/libs/FixedPointInt256.sol
```diff
@@ -0,0 +1,101 @@
...
+
+        return int256(a);
+    }
+
+    /**
+     * @notice convert a signed integer to unsigned integer
+     * @param a int to convert into an unsigned integer.
+     * @return the converted unsigned integer.
+     */
+    function fromInt(int256 a) internal pure returns (uint256) {
```
aparnakr, line 32:
```suggestion
    function IntToUint(int256 a) internal pure returns (uint256) {
```

contracts/libs/FixedPointInt256.sol
```diff
@@ -0,0 +1,101 @@
...
+        if (a < 0) {
+            return uint256(-a);
+        } else {
+            return uint256(a);
```
aparnakr, line 36: does this need an overflow / underflow check?

test/fixedPointInt256.test.ts
```diff
@@ -0,0 +1,112 @@
...
+      const expectedInt = new BigNumber(5)
+
+      assert.equal(
+        (await lib.testFromUint(uint)).toNumber(),
+        expectedInt.toNumber(),
+        'conversion from uint to int mismatch',
+      )
+    })
+
+    it('It should revert converting an unsigned integer greater than uint256(-1) to signed integer', async () => {
```
aparnakr, line 27: does this also work for 2^128 + 1?

haythem96:
> Looks good to me. Should we also have lte, gte functions?

yeah good idea, will add those also.

contracts/libs/FixedPointInt256.sol
haythem96: I don't think so, it is meant to solve this issue https://solidity.readthedocs.io/en/latest/types.html#explicit-conversions

test/fixedPointInt256.test.ts
haythem96: why 2^128 ?!

contracts/libs/FixedPointInt256.sol
antoncoding: agree

contracts/libs/FixedPointInt256.sol
```diff
@@ -0,0 +1,101 @@
...
+
+    /**
+     * @notice multiply two signed integer
+     * @dev rounds to zero if a*b < SCALING_FACTOR / 2
+     * @param a signed integer
+     * @param b signed integer
+     * @return mul of two signed integer
+     */
+    function mul(int256 a, int256 b) internal pure returns (int256) {
+        return (a.mul(b)).add(SCALING_FACTOR / 2) / SCALING_FACTOR;
```
antoncoding, line 68:
```suggestion
        return (a.mul(b)).add(SCALING_FACTOR / 2).div(SCALING_FACTOR);
```

Any reason why not using safemath `.div` here?

haythem96: It is the same, `SCALING_FACTOR` is always equal to 1e18.

test/fixedPointInt256.test.ts
```diff
@@ -0,0 +1,112 @@
...
+
+      assert.equal(
+        (await lib.testFromInt(int)).toNumber(),
+        expectedUint.toNumber(),
+        'conversion from int to uint mismatch',
+      )
+    })
+  })
+
+  describe('Test Addition', () => {
```
antoncoding, line 45: I think we should test more edge cases:
* max positive number + min negative number
* max postitive + max positive
* min negative + min negative

haythem96: for addition, I'm basically using SignedSafeMath from OZ, so those edge cases are already tested.

test/fixedPointInt256.test.ts
haythem96: yes.

test/fixedPointInt256.test.ts
```diff
@@ -0,0 +1,154 @@
...
+    it('Should convert from signed integer to unsigned integer', async () => {
+      const int = new BigNumber(-3)
+      const expectedUint = new BigNumber(3)
+
+      assert.equal(
+        (await lib.testFromInt(int)).toNumber(),
+        expectedUint.toNumber(),
+        'conversion from int to uint mismatch',
+      )
+    })
```
antoncoding, line 42:
```suggestion
    })
    it('Should convert from positive signed integer to unsigned integer', async () => {
      const int = new BigNumber(3)
      const expectedUint = new BigNumber(3)

      assert.equal(
        (await lib.testFromInt(int)).toNumber(),
        expectedUint.toNumber(),
        'conversion from int to uint mismatch',
      )
    })
```

PR #50, opened 2020-07-29
MarginAccount library

haythem96:


contracts/libs/MarginAccount.sol
```diff
@@ -1,12 +1,220 @@
...
+ * SPDX-License-Identifier: UNLICENSED
+ * @dev The MarginAccount is a library that provides Controller with an Account of Vault structs, and
+ * the functions that manipulate vaults. Vaults describe positions that users have.
  */
-// solhint-disable-next-line no-empty-blocks
 library MarginAccount {
+    using SafeMath for uint256;
+
+    /**
+     * @dev Account is a struct corresponding to a user that describes how many vaults that user has.
```
aparnakr, line 14: @haythem96 @antoncoding what do we want to use for consistency? @dev or @notice?

contracts/libs/MarginAccount.sol
```diff
@@ -1,12 +1,220 @@
...
+        Vault storage _vault,
+        address _shortOtoken,
+        uint256 _amount,
+        uint256 _index
+    ) internal {
+        // Valid indexes in any array are between 0 and array.length - 1.
+        // If adding amount into an existant short otoken, make sure _index in the range of 0->length-1
+        if ((_index >= _vault.shortOtokens.length) && ((_index >= _vault.shortAmounts.length))) {
+            _vault.shortOtokens.push(_shortOtoken);
+            _vault.shortAmounts.push(_amount);
```
aparnakr, line 70: so this case is when there is already a valid shortToken?

aparnakr: hmm why is this case needed? Still confused

contracts/libs/MarginAccount.sol
```diff
@@ -1,12 +1,220 @@
...
 
+    /**
+     * @dev remove everything in a vault. Reset short, long and collateral assets and amounts arrays to [].
+     * @param _vault The vault that the user is clearing.
+     */
+    function _clearVault(Vault storage _vault) internal {
+        delete _vault.shortAmounts;
+        delete _vault.longAmounts;
+        delete _vault.collateralAmounts;
+        //TODO: are the following required?
```
aparnakr, line 215: TODO: make a decision on this

test/MarginAccount.test.ts
```diff
@@ -0,0 +1,486 @@
...
+        account = await marginAccountTester.getAccount({from: deployer})
+        assert.equal(account.owner, testAccount.owner, 'MarginAccount.Account owner addr mismatch')
+        assert.equal(account.vaultIds, testAccount.vaultIds, 'Incorrect vaultIds')
+      })
+
+      //TODO: should vaultIds be at 1 or 0 at the end (i.e. how do we do indices)
+      it('should increment vaultID otoken implementation address', async () => {
+        await marginAccountTester.testOpenNewVault({from: deployer})
+        const account = await marginAccountTester.getAccount({from: deployer})
+        assert.equal(account.vaultIds.toString(), '1', 'vaultIds is not 1 after new vault opened')
```
aparnakr, line 80: hmm shouldn't vaultId be 0 when there is 1 vault?

test/MarginAccount.test.ts
```diff
@@ -0,0 +1,486 @@
...
+        const vault = await marginAccountTester.getVault()
+        await marginAccountTester.testClearVault()
+        assert.equal(vault.shortAmounts.length, 0, 'shortAmounts.length should be 0')
+        assert.equal(vault.longAmounts.length, 0, 'longAmounts.length should be 0')
+        assert.equal(vault.collateralAmounts.length, 0, 'collateralAmounts.length should be 0')
+        assert.equal(vault.shortOtokens.length, 0, 'shortOtokens.length should be 0')
+        assert.equal(vault.longOtokens.length, 0, 'longOtokens.length should be 0')
+        assert.equal(vault.collateralAssets.length, 0, 'collateralAssets.length should be 0')
+      })
+    })
```
aparnakr, line 484: Can we do some tests that arbitrarily test combinations of add, remove, clear?

contracts/libs/MarginAccount.sol
zkoticha: yes @antoncoding @haythem96 lmk what you think bout htis. i know haythem you had a completely different proposal entirely

test/MarginAccount.test.ts
zkoticha: that's integration tests no? thot that comes later unless this would also be part of unit testing

antoncoding: yeah IMO those are integration tests

test/MarginAccount.test.ts
antoncoding: I agree we should have a test on vaults[0] before this test case.

contracts/libs/MarginAccount.sol
antoncoding: I think using delete is clean and it payback a little bit gas

test/MarginAccount.test.ts
zkoticha: added tests - lmk if they are what you were thinking of

contracts/libs/MarginAccount.sol
haythem96: We should discuss this in today eng sync, I don't see why we need to clear vault and re-use them, def not gas efficient.

contracts/libs/MarginAccount.sol
```diff
@@ -1,12 +1,220 @@
...
+        uint256 _amount,
+        uint256 _index
+    ) internal {
+        //Check that the removed long token exists in the vault
+        require(
+            (_index < _vault.longOtokens.length) && (_vault.longOtokens[_index] == _longOtoken),
+            "MarginAccount: long otoken address mismatch"
+        );
+
+        _vault.longAmounts[_index] = _vault.longAmounts[_index].sub(_amount);
```
haythem96, line 152: if after removing long, `_vault. longAmounts[_index] == 0`, should we remove the long otoken address from `_vault.longOtokens[_index]` ?! @zkoticha @aparnakr @antoncoding

contracts/libs/MarginAccount.sol
aparnakr: we decided this right? resolving it

contracts/libs/MarginAccount.sol
aparnakr: @haythem96 maybe you can explain?

contracts/libs/MarginAccount.sol
aparnakr, line 215:
```suggestion
```

test/MarginAccount.test.ts
aparnakr: vaultIds represents number of vaults right? In that case it should be 1

test/MarginAccount.test.ts
```diff
@@ -0,0 +1,496 @@
...
+        )
+      })
+
+      //TODO: do we need to check addition overflows?
+      // //TODO: should it revert if trying to add a long otoken to an vault where the same otoken is shorted?'
+      // it('should revert if trying on the wrong vault', async () => {
+      //   // await expectRevert(marginAccountTester.testAddShort(otoken.address, 10, 5, {from: deployer}), 'revert')
+      //   // await expectRevert(await marginAccountTester.testAddShort(Otoken, 10, 3, {from: deployer}), 'revert')
+      //   // await expectRevert(await marginAccountTester.testAddShort(Otoken, 10, 100, {from: deployer}), 'revert')
+      // })
```
aparnakr, line 175: yes this is a good thing to test!

contracts/libs/MarginAccount.sol
haythem96: yeah this is the case when there is no `shortOtokens` => `vault.shortOtokens.length ==0`, or when u are adding a new shortOtoken (e.g there exist one short, but u are adding a new one `vault.shortOtokens.length == 1, _index == 1`).

This case is needed, because when an array is empty `length == 0` the valid indices we can assign into is in the range of `0 and length-1`, that mean if we just do `vault.shortOtokens[index] = address` it will revert with `invalid opcode` message because we are accessing an invalid index., that's why we used `.push`.

Does that make sense ?

haythem96: similar to the case where an `array.length == 2` and we want to add a third element in the index `2`, the available indices we can access using `array[index] = something` are between 0 and `length-1 == 2-1= 1` so in that case will revert, that's why used `.push`

haythem96: In the else block, it basically mean that we are passing an already available index, so we can access without the need of creating a new one using `.push`.

test/MarginAccount.test.ts
haythem96: we are using SafeMath right, why we need to check addition overflows ?

test/MarginAccount.test.ts
aparnakr: well that is reasonable to test for this file imo.

test/MarginAccount.test.ts
aparnakr, line 175:
```suggestion
```

test/MarginAccount.test.ts
```diff
@@ -0,0 +1,489 @@
...
+        const amount = 10
+        await marginAccountTester.testAddCollateral(weth.address, amount, index)
+        const vault = await marginAccountTester.getVault()
+        assert.equal(vault.collateralAmounts[index], new BigNumber(amount))
+      })
+
+      //TODO: understand why below doesn't work
+      // it('should revert if adding weth collateral to wrong index', async () => {
+      //   await expectRevert(marginAccountTester.testAddCollateral(weth.address, 10, 1), "MarginAccount: invalid collateral token position")
+      // })
```
aparnakr, line 188: @antoncoding @haythem96 thoughts on should this be here or controller?

aparnakr, line 188:
```suggestion
```

test/MarginAccount.test.ts
```diff
@@ -0,0 +1,485 @@
...
+          changeAmt.toString(),
+        )
+      })
+
+      //TODO: shouldn't the below revert?
+      // it('should revert if trying to add WETH as collateral to wrong index', async () => {
+      //   const changeAmt = 10
+      //   const index = 1
+      //   await expectRevert(marginAccountTester.testAddCollateral(otoken.address, changeAmt, index), "MarginAccount: invalid collateral token position")
+      // })
```
aparnakr, line 205: @antoncoding @haythem96 should this be here or controller?

test/MarginAccount.test.ts
```diff
@@ -0,0 +1,485 @@
...
+          changeAmt.toString(),
+        )
+      })
+
+      //TODO:  is this out of scope of the margin account?
+      // it('should revert if trying to add otoken as collateral to a vault', async () => {
+      //   await expectRevert(marginAccountTester.testAddCollateral(otoken.address, 10, 3))
+      //   await expectRevert(marginAccountTester.testAddCollateral(otoken.address, 10, 2))
+      //   await expectRevert(marginAccountTester.testAddCollateral(otoken.address, 10, 1))
+      //   await expectRevert(marginAccountTester.testAddCollateral(otoken.address, 10, 0))
```
aparnakr, line 243: can be checked in controller

test/MarginAccount.test.ts
```diff
@@ -0,0 +1,485 @@
...
+          'SafeMath: subtraction overflow',
+        )
+      })
+
+      //TODO: Is this a condition we need?
+      // it('should not be able to remove zero of those short otokens', async () => {
+      //   await marginAccountTester.testRemoveShort(otoken.address, 0, 1)
+      //   const vault = await marginAccountTester.getVault()
+      //   assert.equal(1,2)
+      // })
```
aparnakr, line 303: @antoncoding @haythem96 thoughts?

test/MarginAccount.test.ts
```diff
@@ -0,0 +1,485 @@
...
+
+      it('should not remove short collateral if there is none', async () => {
+        await expectRevert(
+          marginAccountTester.testRemoveCollateral(weth.address, 10, 0),
+          'MarginAccount: collateral token address mismatch',
+        )
+      })
+
+      it('should add more collateral for testing', async () => {
+        await marginAccountTester.testAddShort(weth.address, 10, 0)
```
aparnakr, line 398:
```suggestion
        await marginAccountTester.testAddCollateral(weth.address, 10, 0)
```

test/MarginAccount.test.ts
```diff
@@ -0,0 +1,485 @@
...
+      //   await marginAccountTester.testRemoveShort(otoken.address, 5, 1)
+      //   const vault = await marginAccountTester.getVault()
+      //   assert.equal(1,2)
+      // })
+      //
+      // it('should not be able to remove shortOtokens if there are none remaining', async () => {
+      //   await marginAccountTester.testRemoveShort(otoken.address, 1, 1)
+      //   const vault = await marginAccountTester.getVault()
+      //   assert.equal(1,2)
+      // })
```
aparnakr, line 432: uncomment them!

test/MarginAccount.test.ts
```diff
@@ -0,0 +1,485 @@
...
+      // it('should revert if trying on the wrong vault', async () => {
+      //   // await expectRevert(marginAccountTester.testAddShort(otoken.address, 10, 5, {from: deployer}), 'revert')
+      //   // await expectRevert(await marginAccountTester.testAddShort(Otoken, 10, 3, {from: deployer}), 'revert')
+      //   // await expectRevert(await marginAccountTester.testAddShort(Otoken, 10, 100, {from: deployer}), 'revert')
+      // })
```
aparnakr, line 437: can be dealt with in controller

contracts/libs/MarginAccount.sol
haythem96: @aparnakr @antoncoding

test/MarginAccount.test.ts
haythem96: Can u explain more this test ?

test/MarginAccount.test.ts
haythem96: We are whitelisting WETH as collateral, so that test in controller shouldn't fail.
I think this test mean it should fail to add WETH as collateral in index X if that index have USDC on it ?! if that's the case than yeah this test needed here and should fail.

test/MarginAccount.test.ts
antoncoding: don't need this in the test

test/MarginAccount.test.ts
antoncoding: don't need this IMO

test/MarginAccount.test.ts
antoncoding: I suggest remove weth in this test file, and change this test case to `should revert if trying to add collateral to wrong index`

test/MarginAccount.test.ts
antoncoding: If there's no `amount != 0` constraint in the MarginAccount contract then this is not necessary. But not bad to have it.

test/MarginAccount.test.ts
```diff
@@ -0,0 +1,485 @@
...
+  describe('Otoken Initialization', () => {
+    it('should be able to initialize with put parameter correctly', async () => {
+      await otoken.init(ETH_ADDR, usdc.address, usdc.address, strikePrice, expiry, isPut, {from: deployer})
+      assert.equal(await otoken.underlyingAsset(), ETH_ADDR)
+      assert.equal(await otoken.strikeAsset(), usdc.address)
+      assert.equal(await otoken.collateralAsset(), usdc.address)
+      assert.equal((await otoken.strikePrice()).toString(), strikePrice.toString())
+      assert.equal(await otoken.isPut(), isPut)
+      assert.equal((await otoken.expiry()).toNumber(), expiry)
+    })
```
antoncoding, line 61: 1. Remove the `describe("Otoken Initialization" ()=> {}` block
2. Remove all Otoken related setup / tests. Just use MockERC20 for your `otoken` as in this PR:
https://github.com/opynfinance/GammaProtocol/pull/51/files

contracts/libs/MarginAccount.sol
aparnakr: sure, that seems good

antoncoding (approved):

PR #52, opened 2020-07-30
Task update interfaces

antoncoding:


contracts/interfaces/OtokenInterface.sol
```diff
@@ -2,17 +2,17 @@
 pragma solidity 0.6.10;
 
 interface OtokenInterface {
-    function underlyingAsset() external returns (address);
+    function underlyingAsset() external view returns (address);
```
haythem96, line 5: can we add `get` to the functions names, so it is clear that those functions are getters?

antoncoding: it's not actually a getter function, just the public variables we declared in the Otoken contract.

PR #53, opened 2020-07-30
Oracle module basis

haythem96:
This PR add the basis code for the Oracle module.

PR #54, opened 2020-07-30
Oracle module getter functions

haythem96:
This PR add the getters functions in the Oracle contract.
This PR should be reviewed/merged after merging this pull request #53 .

### DIFF
--- a/contracts/Oracle.sol
+++ b/contracts/Oracle.sol
@@ -4,6 +4,7 @@
 pragma solidity 0.6.10;
 
 import "./interfaces/AggregatorInterface.sol";
+import "./interfaces/AddressBookInterface.sol";
 import "./packages/oz/Ownable.sol";
 
 /**
@@ -12,5 +13,31 @@ import "./packages/oz/Ownable.sol";
  * @notice The Oracle module provide the system with on-chain prices
  */
 contract Oracle is Ownable {
+    /// @dev structure that represent price, and timestamp
+    struct Price {
+        uint256 price;
+        uint256 timestamp; // timestamp at which the price is pulled to this oracle
+    }
 
+    /// @notice AddressBook module
+    address public addressBook;
+
+    /// @dev mapping between oracle and it's locking period
+    mapping(address => uint256) internal oracleLockingPeriod;
+    /// @dev mapping between oracle and it's dispute period
+    mapping(address => uint256) internal oracleDisputePeriod;
+    /// @dev mapping between batch and it's oracle
+    mapping(bytes32 => address) internal batchOracle;
+    /// @dev mapping between batch and it price at specific timestmap
+    mapping(bytes32 => mapping(uint256 => Price)) internal batchPriceAt;
+
+    /**
+     * @dev constructor
+     * @param _addressBook AddressBook module address
+     */
+    constructor(address _addressBook) public {
+        require(_addressBook != address(0), "Invalid address book");
+
+        addressBook = _addressBook;
+    }
 }
--- a/contracts/Oracle.sol
+++ b/contracts/Oracle.sol
@@ -40,7 +40,7 @@ contract Oracle is Ownable {
     }
 
     /**
-     * @notice get batch oracle
+     * @notice get batch oracle. Each underlying-collateral-strike-expiry has its own oracle
      * @param _batch get the price oracle for a specific batch. A batch is the hash of underlying, collateral, strike and expiry.
      * @return oracle address
      */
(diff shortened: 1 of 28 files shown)


contracts/Oracle.sol
```diff
@@ -3,10 +3,68 @@
...
+        uint256 timestamp; // timestamp at which the price is pulled to this oracle
+    }
+
+    /// @dev mapping between oracle and it's locking period
+    mapping(address => uint256) internal oracleLockingPeriod;
+    /// @dev mapping of asset price to it's dispute period
+    mapping(address => uint256) internal oracleDisputePeriod;
+    /// @dev mapping between batch and it's oracle
+    mapping(bytes32 => address) internal batchOracle;
+    /// @dev mapping between batch and it price at specific timestmap
```
aparnakr, line 28: can you explain what a batch is here?

contracts/Oracle.sol
```diff
@@ -3,10 +3,68 @@
...
+    mapping(bytes32 => mapping(uint256 => Price)) internal batchPriceAt;
+
+    /**
+     * @notice get batch price
+     * @param _batch batch hash
```
aparnakr, line 33: explain what a batch is

contracts/Oracle.sol
```diff
@@ -3,10 +3,68 @@
...
+     * @param _timestamp price timestamp
+     * @return price and timestap at which price submitted to this contract
+     */
+    function getBatchPrice(bytes32 _batch, uint256 _timestamp) public view returns (uint256, uint256) {
+        Price memory batchPrice = batchPriceAt[_batch][_timestamp];
+        return (batchPrice.price, batchPrice.timestamp);
+    }
+
+    /**
+     * @notice get batch oracle
```
aparnakr, line 43:
```suggestion
     * @notice get the price oracle for a specific batch. A batch is ... TODO: 
```

contracts/Oracle.sol
```diff
@@ -3,10 +3,68 @@
...
+     * @param _batch batch hash
+     * @return oracle address
+     */
+    function getBatchOracle(bytes32 _batch) public view returns (address) {
+        return batchOracle[_batch];
+    }
+
+    /**
+     * @notice get oracle locking period
```
aparnakr, line 52: what is the oracle locking period? Is it the time up to which an oracle is locked? I'm confused by this

contracts/Oracle.sol
```diff
@@ -3,10 +3,68 @@
...
+     * @dev during an oracle locking period, price can not be submitted to this contract
+     * @param _oracle oracle address
+     * @return locking period
+     */
+    function getOracleLockingPeriod(address _oracle) public view returns (uint256) {
+        return oracleLockingPeriod[_oracle];
+    }
 
+    /**
+     * @notice get oracle dispute period
```
aparnakr, line 62: what is the oracle locking period? Is it after the oracle locking period? Better description needed

contracts/Oracle.sol
haythem96: I explained in the @dev, should I move it to @notice ?

contracts/Oracle.sol
haythem96: I explained in the @dev, should I move it to @notice ?

contracts/Oracle.sol
aparnakr: maybe in notice explain: Each underlying-collateral-strike-expiry has its own oracle

contracts/Oracle.sol
aparnakr: well like what is the locking period? When does it happen? How long is it for? What are the different time periods?

haythem96: cool, for how long and what are different time, there is no specific value, can be any value that the owner set.

antoncoding: Period implies a duration of time, but here we only return `uint256` so it's vague how would that turn into a period. Maybe add something like 
```
@dev time after expiry that the oracle will be locked. During the locking period, price can not be submitted to this contract
```
And for the return pararm:`@return: time in second after expiry`

contracts/Oracle.sol
```diff
@@ -25,6 +25,46 @@ contract Oracle is Ownable {
...
+        return oracleLockingPeriod[_oracle];
+    }
+
+    /**
+     * @notice get oracle dispute period
+     * @dev during an oracle dispute period, the owner of this contract can dispute the submitted price and modify it
+     * @param _oracle oracle address
+     * @return dispute period
+     */
+    function getOracleDisputePeriod(address _oracle) public view returns (uint256) {
```
antoncoding, line 67: Similar to above, add something to explain what the `uint256` means 
```
@dev time after locking period that the owner of this contract can dispute the submitted price and modify it
```

contracts/Oracle.sol
haythem96: cool will add a better description, BTW the return is not time in second after expiry, it is just a period of time, like 15min, 1 hour.. etc, not a specific time.

PR #56, opened 2020-07-31
MarginCalculator Module

antoncoding:
## External functions that this PR implement
* getExpiredCashValue
* getExcessMargin

## Tests
All the test cases for `getExcessMargin` are listed in this [spreadsheet](https://docs.google.com/spreadsheets/d/1jPhQLs8gTt-Trl9aWi15mTp4gsXbjikHwHsiOmxrMss/edit?usp=sharing). Anyone can add some test cases there but please don't change the existing order (id column), otherwise it would be very hard to track and update ;)


contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,198 @@
...
+
+    /**
+     * @notice Return the cash value of an expired oToken.
+     * @dev For call return = Max (0, ETH Price - oToken.strike)
+     * @dev For put return Max(0, oToken.strike - ETH Price)
+     * @param _otoken otoken address
+     */
+    function getExpiredCashValue(address _otoken) public view returns (uint256) {
+        // return 0 if address passed in is 0.
+        if (_otoken == address(0)) return 0;
```
haythem96, line 44: can we revert here ?

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,198 @@
...
+
+    /**
+     * @dev internal function to get underlying price of an otoken.
+     * @param _otoken otoken address
+     */
+    function _getUnerlyingPrice(address _otoken) internal view returns (uint256 price, bool isFinalized) {
+        OtokenInterface otoken = OtokenInterface(_otoken);
+        bytes32 batchId = _getBatchId(otoken);
+        OracleInterface oracle = OracleInterface(AddressBookInterface(addressBook).getOracle());
+        return oracle.getPrice(batchId, otoken.expiryTimestamp());
```
haythem96, line 181: it is `getBatchPrice()` in the Oracle contract not `getPrice()`

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,198 @@
...
+        int256 net = collateralAmount.add(netOtoken);
+        isExcess = net >= 0;
+        netValue = net.intToUint();
+    }
+
+    /**
+     * @dev internal function to get underlying price of an otoken.
+     * @param _otoken otoken address
+     */
+    function _getUnerlyingPrice(address _otoken) internal view returns (uint256 price, bool isFinalized) {
```
haythem96, line 177: typo mistake here `_getUnerlyingPrice` => `_getUnderlyingPrice`

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,198 @@
...
+                 * Put otoken net after expiry: same as above:
+                 *     (long cash value * long amount) - (short cash value * short amount)
+                 */
+                netOtoken = netOtokenAfterExpiry;
+            } else {
+                /*
+                 * Call otoken net after expiry: (net otoken value / underlying price)
+                 */
+                (uint256 underlyingPrice, ) = _getUnerlyingPrice(address(short));
+                // todo: do we need to check isFinalized?
```
haythem96, line 127: To check if price is finalized, we can use the `isPriceFinalized()` from Oracle contract

antoncoding: I'm not sure if we need to check or not.

contracts/MarginCalculator.sol
antoncoding: Fixed

contracts/MarginCalculator.sol
antoncoding: fixed

contracts/MarginCalculator.sol
antoncoding: fixed

contracts/MarginCalculator.sol
antoncoding: What is the expect behavior in this period, (after expiry, before finalized?)

haythem96: I think after expiry before finalized, user should not be able to do anything related to the short option, so I think we should revert if price not finalized ?

antoncoding: Can we do that  (Ban minting and burning after expiry) in the controller? Because this is a view function, if we use revert, then nobody can query this function during that locking period.

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,229 @@
...
+     * @notice Return the cash value of an expired oToken.
+     * @dev For call return = Max (0, ETH Price - oToken.strike)
+     * @dev For put return Max(0, oToken.strike - ETH Price)
+     * @param _otoken otoken address
+     */
+    function getExpiredCashValue(address _otoken) public view returns (uint256) {
+        require(_otoken != address(0), "MarginCalculator: Can't calculate cash value for non-otokens.");
+
+        OtokenInterface otoken = OtokenInterface(_otoken);
+        uint256 strikePrice = otoken.strikePrice();
```
aparnakr, line 46: check that you can't get cash value for non-expired options

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,229 @@
...
+        }
+    }
+
+    /**
+     * @notice return the net value of a vault in either USDC for puts/ ETH for calls
+     * @param _vault the vault that need to be checked
+     * @param _demonimated the token to denominated the result in. Must be the same as short.collateral for now.
+     * @return netValue excess margin or margin requirement amount.
+     * @return isExcess true if there's excess margin in the vault, false if the return amount is margin requirement.
+     */
```
aparnakr, line 63:
```suggestion
    /**
     * @notice returns the net value of a vault in the valid collateral asset for that vault i.e. USDC for puts/ ETH for calls
     * @param _vault the theoretical vault that needs to be checked
     * @param _denominated the token the result is denominated in. Must be the same as short.collateral for now.
     * @return netValue the amount by which the margin is excess margin or margin requirement amount.
     * @return isExcess true if there's excess margin in the vault, false if the return amount is margin requirement.
     */
```

aparnakr: @zkoticha can you also look at the descriptions

aparnakr: Also include the Math equations in @dev

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,229 @@
...
+    /**
+     * @dev internal function that check the long asset is valid for the short asset
+     * @param _vault the vault to check.
+     */
+    function _checkLongAsset(Vault memory _vault) internal view {
+        if (_vault.longOtokens.length == 0 || _vault.shortOtokens.length == 0) return;
+
+        OtokenInterface long = OtokenInterface(_vault.longOtokens[0]);
+        OtokenInterface short = OtokenInterface(_vault.shortOtokens[0]);
+
```
aparnakr, line 191: also check that both oTokens are isValidOTokens?

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,229 @@
...
+                 *     (long cash value * long amount) - (short cash value * short amount)
+                 */
+                netOtoken = netOtokenAfterExpiry;
+            } else {
+                /*
+                 * Call otoken net after expiry: (net otoken value / underlying price)
+                 */
+                (uint256 underlyingPrice, ) = _getUnderlyingPrice(address(short));
+                // todo: do we need to check isFinalized?
+                int256 underlyingPirceInt = FixedPointInt256.uintToInt(underlyingPrice);
```
aparnakr, line 113:
```suggestion
                int256 underlyingPriceInt = FixedPointInt256.uintToInt(underlyingPrice);
```

contracts/MarginCalculator.sol
antoncoding: @haythem96 and I discussed this and we think it would be better to check that in Controller (mint action, which we already did), otherwise the Calculator need to add Whitelist as dependency

contracts/MarginCalculator.sol
antoncoding: That will revert in L48, do we need another check for expiry?
```
require(isFinalized, "MarginCalculator: Oracle price not finalized yet.");
```

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,280 @@
...
+            if (isPut) {
+                netOtoken = _calculateNonExpiredPutNetValue(shortAmount, longAmount, shortStrike, longStrike);
+            } else {
+                netOtoken = _calculateNonExpiredCallNetValue(shortAmount, longAmount, shortStrike, longStrike);
+            }
+        } else {
+            int256 shortCashValue = FixedPointInt256.uintToInt(getExpiredCashValue(short));
+            int256 longCashValue = 0;
+            if (_vault.longOtokens.length > 0)
+                longCashValue = FixedPointInt256.uintToInt(getExpiredCashValue(_vault.longOtokens[0]));
```
zkoticha, line 116: why not just call it getCashValue if it's also used to get cash value prior to expriy?

zkoticha: i.e. cash value is somethign that doesn't change with time, it has no theta

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,280 @@
...
+    }
+
+    /**
+     * @dev calculate non-expired net put spread value.
+     *
+     * Formula: net = (long strike * min (short amount, long amount)) - (short amount * short strike)
+     *
+     * @return net value
+     */
+    function _calculateNonExpiredPutNetValue(
```
zkoticha, line 140: maybe rename to something about difference in strikes or something? or required collat for put spread? @aleone

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,280 @@
...
+    /**
+     * @dev calculate non-expired net call spread value.
+     *                                                min (long amount, short amount) * max (0, long strike - short strike)
+     * net = min(0, long amount - short amount) -  --------------------------------------------------------------------------
+     *                                                                             long strike
+     *
+     * @dev if long strike = 0 (no long token), then return net = short amount.
+     * @return net value
+     */
+    function _calculateNonExpiredCallNetValue(
```
zkoticha, line 158: same comment as before

contracts/MarginCalculator.sol
zkoticha: also does it make a difference if expired or not?

contracts/MarginCalculator.sol
zkoticha: also does it make a difference if expired or not?

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,280 @@
...
+    ) internal pure returns (int256) {
+        return (_longCashValue.mul(_longAmount)).sub(_shortCashValue.mul(_shortAmount)).div(_underlyingPriceInt);
+    }
+
+    /**
+     * @dev internal function that ensure each asset type & amout array have the same length, and length <= 1;
+     * @param _vault the vault to check.
+     */
+    function _checkAssetCount(Vault memory _vault) internal pure {
+        // For the currect version, check lengths of short, long, ollateral <= 1.
```
zkoticha, line 214: "collateral" spelling

zkoticha: "current"

zkoticha: is there a way to check spelling everywhere?

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,280 @@
...
+            _vault.collateralAssets.length == _vault.collateralAmounts.length,
+            "MarginCalculator: Collateral asset and amount mismatch."
+        );
+    }
+
+    /**
+     * @dev internal function that check the long asset is valid for the short asset
+     * @param _vault the vault to check.
+     */
+    function _checkLongAsset(Vault memory _vault) internal view {
```
zkoticha, line 237: longOtoken? or longAsset? (might be confusing as it might be confused for collateral)

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,280 @@
...
+        if (_vault.longOtokens.length == 0 || _vault.shortOtokens.length == 0) return;
+
+        OtokenInterface long = OtokenInterface(_vault.longOtokens[0]);
+        OtokenInterface short = OtokenInterface(_vault.shortOtokens[0]);
+
+        require(long.expiryTimestamp() == short.expiryTimestamp(), "MarginCalculator: Short and Long expiry mismatch.");
```
zkoticha, line 243: yup, keep in mind that this may change in future versions with calendar spreads

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,280 @@
...
+        require(long.strikeAsset() == short.strikeAsset(), "MarginCalculator: Short and Long strike mismatch.");
+        require(
+            long.collateralAsset() == short.collateralAsset(),
+            "MarginCalculator: Short and Long collateral mismatch."
+        );
+    }
+
+    /**
+     * @dev internal function to get underlying price of an otoken.
+     * @param _otoken otoken address
```
zkoticha, line 257: AT return should be here ya?

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,280 @@
...
+        address shortCollateral = OtokenInterface(_vault.shortOtokens[0]).collateralAsset();
+        require(shortCollateral == _demonimated, "MarginCalculator: Denomintated token should be short.collateral");
+
+        int256 netOtoken = _calculateOtokenNetValue(_vault);
+        int256 totalValue = collateralAmount.add(netOtoken);
+        isExcess = totalValue >= 0;
+        netValue = totalValue.intToUint();
+    }
+
+    /**
```
zkoticha, line 89: @param where it at?

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,280 @@
...
+
+    function init(address _addressBook) external initializer {
+        addressBook = _addressBook;
+    }
+
+    /**
+     * @notice Return the cash value of an expired oToken.
+     * @dev For call return = Max (0, ETH Price - oToken.strike)
+     * @dev For put return Max(0, oToken.strike - ETH Price)
+     * @param _otoken otoken address
```
zkoticha, line 40: AT return?

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,280 @@
...
+     * @param _vault the vault to check.
+     */
+    function _checkLongAsset(Vault memory _vault) internal view {
+        if (_vault.longOtokens.length == 0 || _vault.shortOtokens.length == 0) return;
+
+        OtokenInterface long = OtokenInterface(_vault.longOtokens[0]);
+        OtokenInterface short = OtokenInterface(_vault.shortOtokens[0]);
+
+        require(long.expiryTimestamp() == short.expiryTimestamp(), "MarginCalculator: Short and Long expiry mismatch.");
+        require(
```
zkoticha, line 244: why not just check that batchId for both is the same?

zkoticha: remove the 3 require statements for the batchId one

contracts/interfaces/OracleInterface.sol
```diff
@@ -8,7 +8,7 @@ interface OracleInterface {
 
     function isPriceFinalized(bytes32 _batch) external view returns (bool);
 
-    function getPrice(bytes32 _btach, uint256 _timestamp) external view returns (uint256, bool);
+    function getBatchPrice(bytes32 _btach, uint256 _timestamp) external view returns (uint256, bool);
```
zkoticha, line 11: _btach misspelling - should be _batch right

contracts/mocks/MockOracle.sol
```diff
@@ -0,0 +1,26 @@
...
+        addressBook = _addressBook;
+    }
+
+    function getBatchPrice(bytes32 _batch, uint256 _timestamp) external view returns (uint256, bool) {
+        require(_batch != 0x00);
+        require(_timestamp > 0);
+        return (mockedPrice, mockedFinalized);
+    }
+
+    function setMockedStatus(uint256 _pirce, bool _finalized) external {
```
zkoticha, line 22: _price not _pirce

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,280 @@
  * SPDX-License-Identifier: UNLICENSED
  */
 pragma solidity 0.6.10;
+pragma experimental ABIEncoderV2;
+
+import {Initializable} from "./packages/oz/upgradeability/Initializable.sol";
```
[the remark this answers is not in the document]
antoncoding: This line enables using struct as function arguments

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,280 @@
...
+    }
+
+    /**
+     * @notice Return the cash value of an expired oToken.
+     * @dev For call return = Max (0, ETH Price - oToken.strike)
+     * @dev For put return Max(0, oToken.strike - ETH Price)
+     * @param _otoken otoken address
+     */
+    function getExpiredCashValue(address _otoken) public view returns (uint256) {
+        require(_otoken != address(0), "MarginCalculator: Can't calculate cash value for non-otokens.");
```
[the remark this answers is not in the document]
antoncoding: I changed it to `require(_otoken != address(0), "MarginCalculator: Invalid token address.");`

contracts/MarginCalculator.sol
antoncoding: Damn you're smart

antoncoding: Great suggestion! I think It's cleaner. I'll adopt it for now. 

But just want to mention that using hash is gonna be more expensive, and the revert message will be less useful. Happy to have more discussion if anyone has other preference.

antoncoding: Also, in the future version, if we add calendar spread, then we need to check things separately.

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,280 @@
...
+        address short = _vault.shortOtokens[0];
+        int256 shortAmount = _vault.shortAmounts.length > 0 ? FixedPointInt256.uintToInt(_vault.shortAmounts[0]) : 0;
+        int256 longAmount = _vault.longAmounts.length > 0 ? FixedPointInt256.uintToInt(_vault.longAmounts[0]) : 0;
+
+        bool expired = now > OtokenInterface(short).expiryTimestamp();
+        bool isPut = OtokenInterface(short).isPut();
+
+        if (!expired) {
+            int256 shortStrike = FixedPointInt256.uintToInt(OtokenInterface(short).strikePrice());
+            int256 longStrike = 0;
```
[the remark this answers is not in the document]
antoncoding: It works the same in Solidity, so it's more of a coding style concern: 
In most languages, declaration won't assign a default value to a variable, for example in C: 
```c
// in C
int longStrke;
if (condition) {
     longStrike = 5;
}
printf(longStrike)
```
Will crash if condition is not met.

Given this, I think declaration statement and assignments have different rationale, what I want to do here is assigning a value, I think specifying explicitly is clearer.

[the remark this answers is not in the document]
antoncoding: Maybe I can change it from 

```javascript
int256 longStrike = 0;
if (_vault.longOtokens.length > 0)
     longStrike = FixedPointInt256.uintToInt(OtokenInterface(_vault.longOtokens[0]).strikePrice());
```
 To 
```javascript

int256 longStrike = _vault.longOtokens.length > 0 
    ? FixedPointInt256.uintToInt(OtokenInterface(_vault.longOtokens[0]).strikePrice())
    : 0
```

To make the if statement less confusing ;)

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,280 @@
...
+pragma experimental ABIEncoderV2;
+
+import {Initializable} from "./packages/oz/upgradeability/Initializable.sol";
+import {SafeMath} from "./packages/oz/SafeMath.sol";
+import {OtokenInterface} from "./interfaces/OtokenInterface.sol";
+import {OracleInterface} from "./interfaces/OracleInterface.sol";
+import {AddressBookInterface} from "./interfaces/AddressBookInterface.sol";
+import {FixedPointInt256} from "./libs/FixedPointInt256.sol";
+
+// TODO: use this from libs/MarginAccout
```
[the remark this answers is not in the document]
aparnakr: no, merge margin account to dev and pull from dev!

contracts/MarginCalculator.sol
aparnakr: Its better to have that check as well in case there is some issue in oracle where isFinalized & expiry break

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,280 @@
...
+     * @dev For call return = Max (0, ETH Price - oToken.strike)
+     * @dev For put return Max(0, oToken.strike - ETH Price)
+     * @param _otoken otoken address
+     */
+    function getExpiredCashValue(address _otoken) public view returns (uint256) {
+        require(_otoken != address(0), "MarginCalculator: Can't calculate cash value for non-otokens.");
+
+        OtokenInterface otoken = OtokenInterface(_otoken);
+        uint256 strikePrice = otoken.strikePrice();
+        (uint256 underlyingPrice, bool isFinalized) = _getUnderlyingPrice(_otoken);
```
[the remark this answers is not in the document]
antoncoding: yeah

contracts/MarginCalculator.sol
antoncoding: Okay, just added!

contracts/MarginCalculator.sol
antoncoding: I think so, let me find that plugin 🤣

contracts/MarginCalculator.sol
antoncoding: I've added the check. But worth thinking about this view function issue in terms of UX.

contracts/MarginCalculator.sol
antoncoding: This function is only called post expiry. 
But does it make sense to call it `expiredCashValue` ?

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,280 @@
...
+        isExcess = totalValue >= 0;
+        netValue = totalValue.intToUint();
+    }
+
+    /**
+     * @dev Calculate the net value of long token + short token in a vault.
+     * @return netOtoken net worth of long otoken and short otoken
+     */
+    function _calculateOtokenNetValue(Vault memory _vault) internal view returns (int256 netOtoken) {
+        address short = _vault.shortOtokens[0];
```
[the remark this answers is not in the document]
antoncoding: The vault passed here is already checked, that has exactly 1 short asset

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,280 @@
...
+
+        int256 collateralAmount = _vault.collateralAmounts.length > 0
+            ? FixedPointInt256.uintToInt(_vault.collateralAmounts[0])
+            : 0;
+
+        // Vault contains no short tokens: return collateral value.
+        if (_vault.shortOtokens.length == 0) return (collateralAmount.intToUint(), true);
+
+        // For the currenct version, ensure denominated == short.collateral
+        address shortCollateral = OtokenInterface(_vault.shortOtokens[0]).collateralAsset();
```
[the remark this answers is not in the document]
antoncoding: Then it won't pass the first length check.

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,280 @@
...
+     */
+    function getExcessMargin(Vault memory _vault, address _demonimated)
+        public
+        view
+        returns (uint256 netValue, bool isExcess)
+    {
+        _checkAssetCount(_vault);
+        _checkLongAsset(_vault);
+
+        int256 collateralAmount = _vault.collateralAmounts.length > 0
```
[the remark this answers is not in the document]
antoncoding: I think so

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,280 @@
...
+            ? FixedPointInt256.uintToInt(_vault.collateralAmounts[0])
+            : 0;
+
+        // Vault contains no short tokens: return collateral value.
+        if (_vault.shortOtokens.length == 0) return (collateralAmount.intToUint(), true);
```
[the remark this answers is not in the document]
antoncoding: this one only return if length == 0 (no short), and it serves for different purpose than the previous 2 checks.
You can think of the first 2 as requirement checks and they revert if conditions are not matched.

contracts/MarginCalculator.sol
antoncoding: I does not, the computation is similar `( long cash value * long Amount ) - (short cash value * short amount)`, but when we were writing the doc, we decided not to use a single `getCashValue` function for expired and non-expired function.

That's why we ends up having `getExpiredCashValue` for expired otoken, and use `Otoken.strikePrice()` and not use the term `cashValue` for the non-expired ones.

contracts/MarginCalculator.sol
antoncoding: for calls, the formula is different

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,279 @@
...
+    }
+
+    /**
+     * @notice Return the cash value of an expired oToken.
+     * @dev For call return = Max (0, ETH Price - oToken.strike)
+     * @dev For put return Max(0, oToken.strike - ETH Price)
+     * @param _otoken otoken address
+     * @return the cash value of an expired otoken
+     */
+    function getExpiredCashValue(address _otoken) public view returns (uint256) {
```
aparnakr, line 34: We should leave it as getExpiredCashValue, because if not we need to have the oracle price before expiry at each timestamp right? Is our oracle contract designed to do that? If not, better to leave it as is?

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,279 @@
...
+        if (otoken.isPut()) {
+            return strikePrice > underlyingPrice ? strikePrice.sub(underlyingPrice) : 0;
+        } else {
+            return underlyingPrice > strikePrice ? underlyingPrice.sub(strikePrice) : 0;
+        }
+    }
+
+    /**
+     * @notice returns the net value of a vault in the valid collateral asset for that vault i.e. USDC for puts/ ETH for calls
+     * @param _vault _vault the theoretical vault that needs to be checked
```
aparnakr, line 52:
```suggestion
     * @param _vault the theoretical vault that needs to be checked
```

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,279 @@
...
+     * @param _demonimated _denominated the token the result is denominated in. Must be the same as short.collateral for now.
```
aparnakr, line 53:
```suggestion
     * @param _denominated the token the result is denominated in. Must be the same as short.collateral for now.
```

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,279 @@
...
+     * @return netValue the amount by which the margin is above or below the required amount.
+     * @return isExcess true if there's excess margin in the vault. In this case, collateral can be taken out from the vault. False if there is insufficient margin and additional collateral needs to be added to the vault to create the position.
+     */
+    function getExcessMargin(MarginAccount.Vault memory _vault, address _demonimated)
+        public
+        view
+        returns (uint256 netValue, bool isExcess)
+    {
+        // ensure the # asset in collateral, long and short array is valid.
```
aparnakr, line 62:
```suggestion
        // ensure the there is at most 1 asset type used as collateral, at most 1 series of option used as the long option and at most 1 series of option used as the short option. 
```

antoncoding: I think we should have this comment in the description of the `_checkAssetCount` function, not in the comment

contracts/MarginCalculator.sol
antoncoding: Yeah you're right, we don't have the oracle price until expiry.

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,279 @@
...
+        _checkAssetCount(_vault);
+        // ensure the long asset is valid for the short asset.
```
aparnakr, line 64:
```suggestion
        // if there is a short option in the vault, ensure that the long option series being used is valid a margin.
```

antoncoding: Same, I've add this to the function description.

aparnakr: sure, even then, its nice to have that info here.

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,283 @@
...
+        // if netOtoken < 0, the long assets cannot cover the max loss of short assets in the vault.
+        // will need to check if collateral + netOtoken is greater than 0.
+        int256 totalValue = collateralAmount.add(netOtoken);
+        isExcess = totalValue >= 0;
+        netValue = totalValue.intToUint();
+    }
+
+    /**
+     * @notice Calculate the net value of long token + short token in a vault, denominated in collateral asset.
+     * @dev The vault passed in already pass amount array length = asset array length check.
```
aparnakr, line 91: Can we have the formulas for calls vs Puts here? Otherwise its confusing to understand the math.

antoncoding: But we have 4 different formulas, how can we include all of them well in the comment section?

contracts/MarginCalculator.sol
antoncoding: If you can't understand the code without the comment, then the problem is with the  function naming. I think we should rename both functions instead of add duplicate comments.

What are better naming suggestions for `_checkAssetCount` and `_checkLongAsset`?

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,280 @@
...
+        bool isPut = OtokenInterface(short).isPut();
+
+        if (!expired) {
+            int256 shortStrike = FixedPointInt256.uintToInt(OtokenInterface(short).strikePrice());
+            int256 longStrike = 0;
+            if (_vault.longOtokens.length > 0)
+                longStrike = FixedPointInt256.uintToInt(OtokenInterface(_vault.longOtokens[0]).strikePrice());
+
+            if (isPut) {
+                netOtoken = _calculateNonExpiredPutNetValue(shortAmount, longAmount, shortStrike, longStrike);
```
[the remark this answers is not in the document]
antoncoding: getMinMarginRequirement()

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,298 @@
...
+        } else {
+            FixedPointInt256.FixedPointInt memory shortCashValue = _uint256ToFixedPointInt(
+                getExpiredCashValue(address(short))
+            );
+            FixedPointInt256.FixedPointInt memory longCashValue = hasLongInVault
+                ? _uint256ToFixedPointInt(getExpiredCashValue(_vault.longOtokens[0]))
+                : _uint256ToFixedPointInt(0);
+
+            if (isPut) {
+                netOtoken = _calculateExpiredPutNetValue(shortAmount, longAmount, shortCashValue, longCashValue);
```
antoncoding, line 136: expired cash value

contracts/MarginCalculator.sol
aparnakr: _checkIsValidSpread
_checkIsValidNakedOption

aparnakr: _isMaringableLong

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,308 @@
...
+     * @dev The vault passed in already pass amount array length = asset array length check.
+     * @param _vault the theoretical vault that needs to be checked
+     * @return marginRequired the minimal amount of collateral needed in a vault.
+     */
+    function _getMarginRequired(MarginAccount.Vault memory _vault)
+        internal
+        view
+        returns (FixedPointInt256.FixedPointInt memory marginRequired)
+    {
+        // The vault passed in has a short array == 1, so we can just use shortAmounts[0]
```
aparnakr, line 101: mayeb add a require short array length ==1 just so as an independent unit this is fail safe?

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,308 @@
...
+        FixedPointInt256.FixedPointInt memory shortAmount = _uint256ToFixedPointInt(_vault.shortAmounts[0]);
+
+        bool hasLongInVault = _vault.longOtokens.length > 0;
+        FixedPointInt256.FixedPointInt memory longAmount = hasLongInVault
+            ? _uint256ToFixedPointInt(_vault.longAmounts[0])
+            : _uint256ToFixedPointInt(0);
+
+        OtokenInterface short = OtokenInterface(_vault.shortOtokens[0]);
```
aparnakr, line 109: require only 1 long in vault

contracts/MarginCalculator.sol
antoncoding: I don't think internal function has to be treated as an independent unit. 
If we can't write at test that's gonna fail that require statement, then it's redundant, and it's gonna hurt our coverage % as well. 
And even someone can pass in a short array with length 0, the evm is gonna revert anyway.

contracts/MarginCalculator.sol
antoncoding: What does this mean?

contracts/MarginCalculator.sol
antoncoding: Updated!

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,308 @@
...
 
 /**
- *
+ * @title MarginCalculator
+ * @author Opyn
+ * @notice Calculator module that check if a given vault is valid.
  */
-// solhint-disable-next-line no-empty-blocks
-contract MarginCalculator {
+contract MarginCalculator is Initializable {
```
haythem96, line 21: MarginCalculator is not upgradeable module, why inheriting `Initializable` ?

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,308 @@
...
+     * @dev if there is a short option in the vault, ensure that the long option series being used is a valid margin.
+     * @param _vault the vault to check.
+     */
+    function _checkIsMarginableLong(MarginAccount.Vault memory _vault) internal view {
+        if (_vault.longOtokens.length == 0 || _vault.shortOtokens.length == 0) return;
+
+        OtokenInterface long = OtokenInterface(_vault.longOtokens[0]);
+        OtokenInterface short = OtokenInterface(_vault.shortOtokens[0]);
+        require(_getBatchId(long) == _getBatchId(short), "MarginCalculator: Long and short batch mismatch");
+    }
```
haythem96, line 275: In L270 we return (terminate) if no long or short otherwise we use a require statement, will be better to have this function always return true or false for consistency, and have the require statement (return == true) in the `getExcessMargin()`

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,308 @@
...
+            abi.encodePacked(
+                _otoken.underlyingAsset(),
+                _otoken.strikeAsset(),
+                _otoken.collateralAsset(),
+                _otoken.expiryTimestamp()
+            )
+        );
+    }
 
+    function _uint256ToFixedPointInt(uint256 _num) internal pure returns (FixedPointInt256.FixedPointInt memory) {
```
haythem96, line 306: This function does not do any scaling right ?

haythem96: I'm assuming that we are scaling to fixed point with 1e18 in the calculator, but all calculation here does not use that.

haythem96: I think this one should be fixed to:
```
function _uint256ToFixedPointInt(uint256 _num) internal pure returns (FixedPointInt256.FixedPointInt memory) {
        return FixedPointInt256.fromUnscaledInt(SignedConverter.uintToInt(_num));
}
```

haythem96 (changes requested):
`_uint256ToFixedPointInt` is not scaling, should be fixed.

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,308 @@
...
- *
+ * @title MarginCalculator
+ * @author Opyn
+ * @notice Calculator module that check if a given vault is valid.
  */
-// solhint-disable-next-line no-empty-blocks
-contract MarginCalculator {
+contract MarginCalculator is Initializable {
+    using SafeMath for uint256;
+    using FixedPointInt256 for FixedPointInt256.FixedPointInt;
```
haythem96, line 23: Also add here 
`using FixedPointInt256 for int256;`

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,308 @@
...
+     * @notice returns the net value of a vault in the valid collateral asset for that vault i.e. USDC for puts/ ETH for calls
+     * @param _vault the theoretical vault that needs to be checked
+     * @param _denominated the token the result is denominated in. Must be the same as short.collateral for now.
+     * @return netValue the amount by which the margin is above or below the required amount.
+     * @return isExcess true if there's excess margin in the vault. In this case, collateral can be taken out from the vault. False if there is insufficient margin and additional collateral needs to be added to the vault to create the position.
+     */
+    function getExcessMargin(MarginAccount.Vault memory _vault, address _denominated)
+        public
+        view
+        returns (uint256 netValue, bool isExcess)
```
haythem96, line 63: Remove variable names and assign them + return inside function.

contracts/MarginCalculator.sol
haythem96: Also we mostly need a function `toUnscaledInt` that scale down before using `intToUint` in any return statement.

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,308 @@
...
+        FixedPointInt256.FixedPointInt memory collateralAmount = _vault.collateralAmounts.length > 0
+            ? _uint256ToFixedPointInt(_vault.collateralAmounts[0])
+            : _uint256ToFixedPointInt(0);
+
+        // Vault contains no short tokens: return collateral value.
+        if (_vault.shortOtokens.length == 0) return (SignedConverter.intToUint(collateralAmount.value), true);
+
+        // For the currenct version, ensure denominated == short.collateral
+        address shortCollateral = OtokenInterface(_vault.shortOtokens[0]).collateralAsset();
+        require(shortCollateral == _denominated, "MarginCalculator: Denomintated token should be short.collateral");
```
haythem96, line 80: change require message to "MarginCalculator: Denomintated token should be the short otoken's collateral"

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,308 @@
...
+    {
+        // ensure the number of collateral, long and short array is valid.
+        _checkIsValidSpread(_vault);
+        // ensure the long asset is valid for the short asset.
+        _checkIsMarginableLong(_vault);
+
+        // collateral amount is always positive.
+        FixedPointInt256.FixedPointInt memory collateralAmount = _vault.collateralAmounts.length > 0
+            ? _uint256ToFixedPointInt(_vault.collateralAmounts[0])
+            : _uint256ToFixedPointInt(0);
```
haythem96, line 73: Should check collateral amount using collateral asset array.
e.g my vault.collateralAmounts.length=0, I deposit 100USDC, vault.collateralAmounts.length=1, I withdraw 100USDC, still vault.collateralAmounts.length=1. In this case it is the same end result `_uint256ToFixedPointInt(0)` but will be better to have consistency on how to know wether a vault have assets or not.

Another example is L76 where it make a huge impact.

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,308 @@
...
+
+        // Vault contains no short tokens: return collateral value.
+        if (_vault.shortOtokens.length == 0) return (SignedConverter.intToUint(collateralAmount.value), true);
```
haythem96, line 76: If I mint 100 short, I burn them, _vault.shortOtokens.length will be 1 (_vault.shortOtokens[0] will be equal to address[0]), here it should be changed to `if (_vault.shortOtokens.length == 0 || _vault.shortOtokens[0] == address(0))

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,308 @@
...
+     */
+    function _getMarginRequired(MarginAccount.Vault memory _vault)
+        internal
+        view
+        returns (FixedPointInt256.FixedPointInt memory marginRequired)
+    {
+        // The vault passed in has a short array == 1, so we can just use shortAmounts[0]
+        FixedPointInt256.FixedPointInt memory shortAmount = _uint256ToFixedPointInt(_vault.shortAmounts[0]);
+
+        bool hasLongInVault = _vault.longOtokens.length > 0;
```
haythem96, line 104: should be `bool hasLongInVault = (_vault.longOtokens.length > 0) && (_vault.longOtokens.[0] != address(0));`

contracts/MarginCalculator.sol
antoncoding: now in the address book it's using updateImpl for calculator, implying that it's upgradable right?

contracts/MarginCalculator.sol
antoncoding: Resolve this now, will open another issue for this

contracts/MarginCalculator.sol
haythem96: Then we should change that in another PR.

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,319 @@
...
+            "MarginCalculator: Collateral asset and amount mismatch."
+        );
+    }
+
+    /**
+     * @dev if there is a short option in the vault, ensure that the long option series being used is a valid margin.
+     * @param _vault the vault to check.
+     */
+    function _isMarginableLong(MarginAccount.Vault memory _vault) internal view returns (bool) {
+        if (_vault.longOtokens.length == 0 || _vault.shortOtokens.length == 0) return true;
```
haythem96, line 276: this one also should be fixed
`if(
  (_vault.longOtokens.length == 0 || _vault.longOtokens[0] == address(0)) ||
  (_vault. shortOtokens.length == 0 || _vault. shortOtokens[0] == address(0))
) {
 return true
}

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,319 @@
...
+    /**
+     * @notice Calculate the amount of collateral needed for a spread vault.
+     * @dev The vault passed in already pass amount array length = asset array length check.
+     * @param _vault the theoretical vault that needs to be checked
+     * @return marginRequired the minimal amount of collateral needed in a vault.
+     */
+    function _getMarginRequired(MarginAccount.Vault memory _vault)
+        internal
+        view
+        returns (FixedPointInt256.FixedPointInt memory marginRequired)
```
haythem96, line 105: Will be better to remove the variable assignment here `marginRequired` and do it inside function + `return`, I think it is more clear and easy to find the return statement.

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,319 @@
...
+    }
+
+    /**
+     * @notice returns the net value of a vault in the valid collateral asset for that vault i.e. USDC for puts/ ETH for calls
+     * @param _vault the theoretical vault that needs to be checked
+     * @param _denominated the token the result is denominated in. Must be the same as short.collateral for now.
+     * @return netValue the amount by which the margin is above or below the required amount.
+     * @return isExcess true if there's excess margin in the vault. In this case, collateral can be taken out from the vault. False if there is insufficient margin and additional collateral needs to be added to the vault to create the position.
+     */
+    function getExcessMargin(MarginAccount.Vault memory _vault, address _denominated)
```
haythem96, line 60: getExcessMargin() should get only the vault as argument, right ? in that case we will get rid of the if/else statement in the Controller to check the short array length, `_denominated` can be taken directly from the vault.

haythem96: https://github.com/opynfinance/GammaProtocol/blob/4037eb528760adb16b4f108123f82079f4b1def4/contracts/Controller.sol#L259

contracts/MarginCalculator.sol
```diff
@@ -2,11 +2,319 @@
...
+        // return (otoken.expiryTimestamp(), true);
+        return oracle.getExpiryPrice(otoken.underlyingAsset(), otoken.expiryTimestamp());
+    }
+
+    /**
+     * @dev internal function to hash paramters and get batch id. Same batch id means same product with same expiry.
+     * @param _otoken otoken address
+     * @return id the batchDd of an otoken
+     */
+    function _getBatchId(OtokenInterface _otoken) internal view returns (bytes32 id) {
```
haythem96, line 306: Remove this.

haythem96 (approved):
great work.

PR #57, opened 2020-07-31
Oracle setter functions

haythem96:
This PR add the code to set the following parameters:
- Batch oracle address
- Oracle locking period
- Oracle dispute period

Should be reviewed/merged after merging #54 .

PR #58, opened 2020-07-31
Oracle module: check locking and dispute period

haythem96:
This PR add the functions to check if a locking or dispute period is over.
Should be reviewed/merged after #57 .

### DIFF
--- a/contracts/Oracle.sol
+++ b/contracts/Oracle.sol
@@ -4,6 +4,7 @@
 pragma solidity 0.6.10;
 
 import "./interfaces/AggregatorInterface.sol";
+import "./interfaces/AddressBookInterface.sol";
 import "./packages/oz/Ownable.sol";
 
 /**
@@ -12,5 +13,31 @@ import "./packages/oz/Ownable.sol";
  * @notice The Oracle module provide the system with on-chain prices
  */
 contract Oracle is Ownable {
+    /// @dev structure that represent price, and timestamp
+    struct Price {
+        uint256 price;
+        uint256 timestamp; // timestamp at which the price is pulled to this oracle
+    }
 
+    /// @notice AddressBook module
+    address public addressBook;
+
+    /// @dev mapping between oracle and it's locking period
+    mapping(address => uint256) internal oracleLockingPeriod;
+    /// @dev mapping between oracle and it's dispute period
+    mapping(address => uint256) internal oracleDisputePeriod;
+    /// @dev mapping between batch and it's oracle
+    mapping(bytes32 => address) internal batchOracle;
+    /// @dev mapping between batch and it price at specific timestmap
+    mapping(bytes32 => mapping(uint256 => Price)) internal batchPriceAt;
+
+    /**
+     * @dev constructor
+     * @param _addressBook AddressBook module address
+     */
+    constructor(address _addressBook) public {
+        require(_addressBook != address(0), "Invalid address book");
+
+        addressBook = _addressBook;
+    }
 }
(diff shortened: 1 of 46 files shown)


contracts/Oracle.sol
```diff
@@ -3,10 +3,133 @@
...
+    }
+
+    /**
+     * @notice set oracle locking period
+     * @dev can only be called by owner
+     * @param _oracle oracle address
+     * @param _lockingPeriod locking period
+     */
+    function setLockingPeriod(address _oracle, uint256 _lockingPeriod) external onlyOwner {
+        oracleLockingPeriod[_oracle] = _lockingPeriod;
```
antoncoding, line 123: Should we have events for this and `setDisputePeriod`?

test/oracle.test.ts
```diff
@@ -0,0 +1,84 @@
+import {ethers} from 'ethers'
```
antoncoding, line 1: where do you use it?

test/oracle.test.ts
```diff
@@ -0,0 +1,84 @@
...
+        'oracle dispute period mismatch',
+      )
+    })
+
+    it('should check if locking period is over when price timestmap equal to zero', async () => {
+      const isOver = await oracle.isDisputePeriodOver(batchOracle, await time.latest(), {from: owner})
+      const expectedResult = false
+      assert.equal(isOver, expectedResult, 'dispute period check mismatch')
+    })
+  })
```
antoncoding, line 83: Add test for: 
* batchPrice.timestamp != 0 (https://coveralls.io/builds/32468982/source?filename=contracts/Oracle.sol#L99)

contracts/Oracle.sol
```diff
@@ -3,10 +3,133 @@
...
+    /**
+     * @notice get batch price
+     * @param _batch batch hash
+     * @param _timestamp price timestamp
+     * @return price and timestap at which price submitted to this contract
+     */
+    function getBatchPrice(bytes32 _batch, uint256 _timestamp) public view returns (uint256, uint256) {
+        Price memory batchPrice = batchPriceAt[_batch][_timestamp];
+        return (batchPrice.price, batchPrice.timestamp);
+    }
```
antoncoding, line 43: Remove this function if we don't have setter and tests in this PR.

contracts/Oracle.sol
haythem96: yeah will add events in the setter PR.

test/oracle.test.ts
haythem96: Will add it in another PR when I add the set price function.

contracts/Oracle.sol
haythem96: This one was added in the getters PR, I think we can keep it as I will add tests for it in the next PR.

test/oracle.test.ts
haythem96: Yeah I installed it and didn't use it, was playing with it to mimic `keccak256(abi.encode())` hahsing as web3js is not great at that... I removed it for now.

contracts/Oracle.sol
```diff
@@ -67,4 +77,72 @@ contract Oracle is Ownable {
     function getOracleDisputePeriod(address _oracle) public view returns (uint256) {
         return oracleDisputePeriod[_oracle];
     }
+
+    /**
+     * @notice check if locking period is over
+     * @param _batch batch hash
```
aparnakr, line 83: expand on what batch is

PR #60, opened 2020-08-01
Remove WETH from AddressBook

antoncoding:


antoncoding:
> Can weth be added to whitelist module?

why do we need that in whitelist module? Shouldn't we just whitelist WETH with `whitelistCollateral` as part of the setup process?

haythem96:
> > Can weth be added to whitelist module?
> 
> why do we need that in whitelist module? Shouldn't we just whitelist WETH with `whitelistCollateral` as part of the setup process?

The Controller still need to know the exact WETH address, to check at every call if asset address equal to WETH and do the wrapping and unwrapping (deposit, withdraw).
That's why I thought we put the WETH address in AddressBook, cause here we are storing a contract address also, which is the wrapper contract, not an asset address... we will anyway still whitelist in the Whitelist module. CC @aparnakr

aparnakr:
> > > Can weth be added to whitelist module?
> > 
> > 
> > why do we need that in whitelist module? Shouldn't we just whitelist WETH with `whitelistCollateral` as part of the setup process?
> 
> The Controller still need to know the exact WETH address, to check at every call if the asset address equal to WETH and do the wrapping and unwrapping (deposit, withdraw).
> That's why I thought we put the WETH address in AddressBook, cause here we are storing a contract address also, which is the wrapper contract, not an asset address... we will anyway still whitelist in the Whitelist module. CC @aparnakr

No strong thoughts. Discussed this with haythem. The only think I agree we need is the WETH address. If we think of WETH as a wrapper, addressbook makes sense. If we think of WETH as collateral / asset, whitelist module makes more sense.

antoncoding:
@aparnakr @haythem96 let's close it for now because we all agree that putting it in the AddressBook make sense. We can re-open this if we change our mind

PR #61, opened 2020-08-01
Fix compiler warnings

antoncoding:
* Use `// comment` instead of `/// @notice` in struct.

### DIFF
--- a/contracts/libs/Actions.sol
+++ b/contracts/libs/Actions.sol
@@ -23,65 +23,65 @@ library Actions {
     }
 
     struct ActionArgs {
-        /// @notice The type of action that is being performed on the system
+        // The type of action that is being performed on the system
         ActionType actionType;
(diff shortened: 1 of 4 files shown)


contracts/libs/Actions.sol
```diff
@@ -23,65 +23,65 @@ library Actions {
...
-        /// @notice The address of the account owner
+        // The address of the account owner
         address owner;
     }
 
     struct DepositArgs {
-        /// @notice The address of the account owner
+        // The address of the account owner
         address owner;
-        /// @notice The index of the vault to which the asset will be added
```
aparnakr, line 55: @antoncoding should these be @dev?

antoncoding: Compiler will complain when we use either @notice or @dev in struct. Those are meant for state variables.

PR #72, opened 2020-08-14
feat/Add multiple collateral transfer to user and pool.

aleone:
Adds batchTransferToUser and batchTransferToPool functions.  Extend testing file for testing multiple collateral transfers.  Fixed a few typos.

There are difficulties testing overloaded functions in typescript, so we decidede to use different function names for multiple vs single transfers.

**Small changes to spec, which have been updated:**
name of multiple transfer functions changed
multiple transfer functions do not return anything
scaling happens in controller now

### DIFF
--- a/contracts/MarginPool.sol
+++ b/contracts/MarginPool.sol
@@ -47,7 +47,6 @@ contract MarginPool {
      * @notice transfers asset from user to pool
      * @dev all tokens are scaled to have 1e18 precision in contracts,
      *      so amounts are scaled down to native token decimals using _calcTransferAmount().
-     *      If _asset equal to WETH address, transfer WETH from Controller address to pool
      * @param _asset address of asset to transfer
      * @param _user address of user to transfer assets from
      * @param _amount amount of token to transfer from _user, scaled to 1e18 of precision
@@ -57,8 +56,8 @@ contract MarginPool {
         address _asset,
         address _user,
         uint256 _amount
-    ) external onlyController returns (bool) {
-        require(_amount > 0, "MarginPool: transferToPool amount is below 0");
+    ) public onlyController returns (bool) {
+        require(_amount > 0, "MarginPool: transferToPool amount is below or equal to 0");
 
         // transfer val from _user to pool
         return ERC20Interface(_asset).transferFrom(_user, address(this), _amount);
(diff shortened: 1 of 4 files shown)


contracts/MarginPool.sol
```diff
@@ -77,10 +76,58 @@ contract MarginPool {
...
+     *      token decimals in Controller before being passed to MarginPool
+     * @param _asset addresses of assets to transfer
+     * @param _user addresses of users to transfer assets to
+     * @param _amount amount of each token to transfer to _user, scaled to 1e18 of precision
+     */
+    function batchTransferToPool(
+        address[] memory _asset,
+        address[] memory _user,
+        uint256[] memory _amount
+    ) public onlyController {
```
antoncoding, line 98:
```suggestion
    ) external onlyController {
```
Looks like it can be declared as external

contracts/MarginPool.sol
```diff
@@ -77,10 +76,58 @@ contract MarginPool {
...
+     *      token decimals in Controller before being passed to MarginPool
+     * @param _asset addresses of assets to transfer
+     * @param _user addresses of users to transfer assets to
+     * @param _amount amount of each token to transfer to _user, scaled to 1e18 of precision
+     */
+    function batchTransferToUser(
+        address[] memory _asset,
+        address payable[] memory _user,
+        uint256[] memory _amount
+    ) public onlyController {
```
antoncoding, line 122:
```suggestion
    ) external onlyController {
```

contracts/MarginPool.sol
```diff
@@ -77,10 +76,58 @@ contract MarginPool {
...
+        require(
+            _asset.length == _user.length && _user.length == _amount.length,
+            "MarginPool: batchTransferToPool array lengths are not equal"
+        );
+
+        for (uint256 i = 0; i < _asset.length; i++) {
+            // transfer val from _user to pool
+            transferToPool(_asset[i], _user[i], _amount[i]);
```
antoncoding, line 106: In standard ERC20 `transfer` and `transferFrom`, the returned boolean will indicate whether the transfer is successful (ERC20 doesn't require a revert in failed transactions, although it's widely used). 
So we shouldn't ignore the returned value from `transferToUser` and `transferToPool`.
```suggestion
            require(transferToPool(_asset[i], _user[i], _amount[i]), "MarginPool: Transfer to pool failed");
```

contracts/MarginPool.sol
```diff
@@ -77,10 +76,58 @@ contract MarginPool {
...
+        uint256[] memory _amount
+    ) public onlyController {
+        require(
+            _asset.length == _user.length && _user.length == _amount.length,
+            "MarginPool: batchTransferToUser array lengths are not equal"
+        );
+
+        for (uint256 i = 0; i < _asset.length; i++) {
+            // transfer val from Pool to _pool
+            transferToUser(_asset[i], _user[i], _amount[i]);
```
antoncoding, line 130:
```suggestion
            require(transferToUser(_asset[i], _user[i], _amount[i]), "MarginPool: TransferFrom failed");
```

aleone:
> Adding the above `require` suggested above will require more some tests on those dumber ERC20 instances.
> 
> We will need a MockDumbERC20 contract, that return false on transfer, transferFrom, approve, so we can use it to test this behavior.

Should I just create/clone an ERC-20 contract in mocks, change the code to enable the mentioned test, and add it to the testing file?

The behavior that we want is a revert, due to our require statements if any ERC-20 returns false for any of those functions, correct?

antoncoding:
> 
> Should I just create/clone an ERC-20 contract in mocks, change the code to enable the mentioned test, and add it to the testing file?

Yes! You can create a `MockERC20Error`, change the function to return directly and then use it in the test.
 
> The behavior that we want is a revert, due to our require statements if any ERC-20 returns false for any of those functions, correct?

I also think that `transferToUser` and `batchTransferToUser` should have the same return format. If we're reverting when the ERC20 operation failed, we should just revert in `tranferToUser` instead of letting it revert false. Then `batchTransferToUser` will just be a loop without requirement statements.
@haythem96 want to get your thoughts for this.

haythem96:
> > Should I just create/clone an ERC-20 contract in mocks, change the code to enable the mentioned test, and add it to the testing file?
> 
> Yes! You can create a `MockERC20Error`, change the function to return directly and then use it in the test.
> 
> > The behavior that we want is a revert, due to our require statements if any ERC-20 returns false for any of those functions, correct?
> 
> I also think that `transferToUser` and `batchTransferToUser` should have the same return format. If we're reverting when the ERC20 operation failed, we should just revert in `tranferToUser` instead of letting it revert false. Then `batchTransferToUser` will just be a loop without requirement statements.
> @haythem96 want to get your thoughts for this.

yeah for the single transfer, was just thinking to follow same ERC20 standard and return a bool...
Not strong opinion re this, happy with either one.

contracts/MarginPool.sol
```diff
@@ -57,17 +56,17 @@ contract MarginPool {
         address _asset,
         address _user,
         uint256 _amount
-    ) external onlyController returns (bool) {
-        require(_amount > 0, "MarginPool: transferToPool amount is below 0");
+    ) public onlyController returns (bool) {
```
aparnakr, line 59: why public? Why not leave it as external?

PR #75, opened 2020-08-19
Controller skeleton code

zkoticha:
- Controller module skeleton
- Set operator function
- Pause system function


contracts/Controller.sol
```diff
@@ -1,12 +1,297 @@
...
+
+    mapping (address  => MarginAccount.Account) internal accounts;
+    mapping (address  => mapping (uint256  => MarginAccount.Vault)) internal vaults;
+    mapping (address => mapping(address => bool)) internal operators;
+
+    event NewOperatorSet(address newOperator);
+
+    modifier isPaused {
+        require(
+            throw
```
antoncoding, line 29: remove this `throw` otherwise it can't be compiled

haythem96: yeah I think for now remove all the code inside the modifier, just keep the modifier declaration.

contracts/Controller.sol
```diff
@@ -1,12 +1,297 @@
...
+        //    setProductUnderlyingPrice (address _otoken, uint256 _roundsBack, ) external
+        //    Get oToken.strike, oToken.underlying
+    }
 
-}
+    /**
+    * @dev the oracle sets the price of the underlying of an otoken.
+    //TODO: why are we setting price of individual otokens, shouldn't we set for an entire batch at once?
+    */
+    function setProductUnderlyingPrice (address _otoken, uint256 _roundsBack, ) external {
```
haythem96, line 286: Remove the last `,`

contracts/Controller.sol
```diff
@@ -1,12 +1,297 @@
...
+    //TODO: IS THERE SUPPOSED TO BE ANOTHER param?
+        //TODO: why are we setting price of individual otokens, shouldn't we set for an entire batch at once?
```
haythem96, line 288: We are setting price per batch here, we just pass Otoken as argument, to hash and get the batch and use that.

contracts/Controller.sol
```diff
@@ -1,12 +1,217 @@
...
+     * @param _args MintArgs structure
+     */
+    function _mintOtoken(Actions.MintArgs memory _args) internal isAuthorized(_args.owner) {}
+
+    /**
+     * @notice burn option
+     * @dev only account owner or operator can withdraw long option from vault
+     * @param _args MintArgs structure
+     */
+    function _burnOtoken(Actions.BurnArgs memory _args) internal {}
```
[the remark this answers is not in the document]
antoncoding: yes, this is one of the operation just like everything else

antoncoding:
Also comment out the modifier, otherwise the coverage test is not gonna pass

antoncoding (approved):

PR #76, opened 2020-08-20
Setup echidna

aparnakr:
High Level:
This PR sets up the environment to write echidna tests. 

Specific Changes:
Created a unit tests folder
Created an echidna folder


contracts/echidna/mocks/EchidnaMockAddressBook.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity 0.6.10;
+
+contract EchidnaMockAddressBook {
+    constructor() public {}
+
+    function getController() external view returns (address) {
+        return 0x00a329C0648769a73afAC7F9381e08fb43DBEA70;
```
antoncoding, line 8: Will hardcoding this cause any problem?

aparnakr: it is hardcoded to return the owner address for now, which it thinks is the controller. This is to test the "onlyController" functionality.

antoncoding (approved):

contracts/echidna/mocks/EchidnaMockAddressBook.sol
antoncoding: Got it I see it now. I misunderstood it as a hardcoded accounts value that might be different if we run truffle with different seed phrase 😹

haythem96 (commented):
Can we remove `PropertiesOtokenTransferable` ? 
Also why `/tests/unit-test/` ?

antoncoding (approved):

antoncoding:
Reopen a new PR without touching unit tests

PR #77, opened 2020-08-20
Crytic cleanup

antoncoding:
## Fix
* [Github PR Template bug](https://stackoverflow.com/questions/52139192/github-pull-requests-template-not-showing): Can't use multiple PR template on Github now, so I remove the intermediate folder, after merging this, we should have PR template for new PRs!

## Crytic Config
* Ignore all low severity checks
* Ignore all informational checks
* Ignore all optimization checks


contracts/AddressBook.sol
```diff
@@ -224,9 +224,9 @@ contract AddressBook is Ownable {
 
         if (proxyAddress == address(0)) {
             OwnedUpgradeabilityProxy proxy = new OwnedUpgradeabilityProxy();
-            proxy.upgradeToAndCall(_newAddress, params);
             setAddress(_id, address(proxy));
             emit ProxyCreated(_id, address(proxy));
+            proxy.upgradeToAndCall(_newAddress, params);
```
haythem96, line 229: why changing this here ?

contracts/Oracle.sol
```diff
@@ -195,7 +195,7 @@ contract Oracle is Ownable {
 
         bool iterate = true;
         uint256 roundBack = _roundsBack;
-        uint256 price;
+        uint256 price = 0;
```
haythem96, line 198: by default uint in solidity are initialized to 0, idk why crytic show error here 😂

haythem96 (approved):

contracts/AddressBook.sol
antoncoding: it's a better pattern to update storage data and then make external calls

haythem96: yeah to avoid reentrancy attack (y)

PR #79, opened 2020-08-20
Remove owner address from account structure

haythem96:


antoncoding:
Can we remove the account structure?

antoncoding (approved):

contracts/libs/MarginAccount.sol
```diff
@@ -12,8 +12,6 @@ library MarginAccount {
 
     // Account is a struct corresponding to a user that describes how many vaults that user has.
     struct Account {
-        // the owner whose vaults we are looking for
-        address owner;
         // the number of vaults a user has, starting at index 0 and incrementing. vaultCounter sorted chronologically.
         uint256 vaultCounter;
     }
```
aparnakr, line 17: do we even need this structure in that case?

haythem96: yeah I think better to store the vault counter here.

PR #81, opened 2020-08-20
Add contract size check in CircleCI

antoncoding:
# Task: Add contract size check in CircleCI

## High Level Description
Make sure no individual contract size is > 48 KB.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,8 +29,11 @@ jobs:
           name: Deploy Contracts
           command: npm run deploy:development
       - run:
+          name: Test Contract Size
+          command: npm run deploy:development
+      - run:
           name: Test Contracts
-          command: npm run test
+          command: npm run contract-size
       - run: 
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -2,42 +2,120 @@ version: 2.1
 orbs:
   coveralls: coveralls/coveralls@1.0.6
   node: circleci/node@1.1.6
+
 workflows:
-  env-vars:
+  build-test-report:
     jobs:
```
haythem96, line 8: I think we can change the order here to lint, size-check, build, test and then coverage.

antoncoding: size-check depends on build because it needs to use the compiled files in `build/`, and I'm not sure what you mean by changing the order to that. You mean the `build-test-report` string?

haythem96: do those jobs start all at the same time ?

PR #84, opened 2020-08-24
Update fixed point library

### LINKED ISSUE #78: Use struct instead of uint256 for fixed point library
- Move `uintToInt()` and `intToUint()` to another library
- Use struct to represent fixed point variables

haythem96:
# Update fixed point library

## High Level Description

- Update fixed point library to use struct to represent scaled number instead of a simple int variable
- Moved `uintToInt()` and `intToUint()` to a new `SignedConverter` library because they are not related to FixedPoint lib.

Closes #78 .

### Code

- [X] Unit test 100% coverage

### Documentation

- [X] Have you added your tests to the testing doc?

### DIFF
--- a/test/signedConverter.test.ts
+++ b/test/signedConverter.test.ts
@@ -1,0 +1,54 @@
+import {SignedConverterTesterInstance} from '../build/types/truffle-types'
+import BigNumber from 'bignumber.js'
+
+const {expectRevert} = require('@openzeppelin/test-helpers')
+
+const SignedConverterTester = artifacts.require('SignedConverterTester.sol')
+
+contract('FixedPointInt256 lib', () => {
+  let lib: SignedConverterTesterInstance
+
+  before('set up contracts', async () => {
+    lib = await SignedConverterTester.new()
+  })
+
+  describe('Test type conversion', () => {
+    it('Should convert from unsigned integer to signed integer', async () => {
+      const uint = new BigNumber(5)
+      const expectedInt = new BigNumber(5)
+
+      assert.equal(
+        (await lib.testFromUint(uint)).toNumber(),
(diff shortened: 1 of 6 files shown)


test/signedConverter.test.ts
```diff
@@ -0,0 +1,54 @@
...
+    it('Should convert from signed integer to unsigned integer', async () => {
+      const int = new BigNumber(-3)
+      const expectedUint = new BigNumber(3)
+
+      assert.equal(
+        (await lib.testFromInt(int)).toNumber(),
+        expectedUint.toNumber(),
+        'conversion from int to uint mismatch',
+      )
+    })
```
aparnakr, line 42: can we maybe have the conversion from signed to unsigned error for -ve numbers and have a separate absolute value function? That way makes it super clear and makes sure we don't mess up.

test/signedConverter.test.ts
```diff
@@ -0,0 +1,54 @@
...
+
+      assert.equal(
+        (await lib.testFromUint(uint)).toNumber(),
+        expectedInt.toNumber(),
+        'conversion from uint to int mismatch',
+      )
+    })
+
+    it('It should revert converting an unsigned integer greater than uint256(-1) to signed integer', async () => {
+      const uint = new BigNumber(2).multipliedBy(1e256).minus(1)
```
antoncoding, line 28: Isn't this `2*10^256 -1`? and uint256(-1) should be `2^256 - 1` 🤔

antoncoding (approved):

antoncoding:
> Doesn't this also affect the way margin calculator is implemented?

yes

antoncoding:
Merging now. Opened another issue for abs() #85

PR #89, opened 2020-08-25
Controller module: getters and view functions

haythem96:
# Controller module: getters/view functions

## High Level Description

This PR implement controller module modifiers, getters and view function:
- modifier isNotPaused
- modifier isAuthorized
- isOperator()
- getVaultBalances()
- isPriceFinalized()
- isExpired()
- getVault()

Should be reviewed/merged after #75 .

### Code

- [ ] Unit test 100% coverage (will be back at 100% the coming PR)

### Documentation

- [X] Have you added your tests to the testing doc?


contracts/Controller.sol
```diff
@@ -47,25 +51,27 @@ contract Controller is Ownable {
...
+     * @param _accountOwner account owner address
      */
-    // modifier isAuthorized(address _sender) {
-    //     _;
-    // }
+    modifier isAuthorized(address _sender, address _accountOwner) {
+        MarginAccount.Account memory senderAccount = accounts[_sender];
+
+        require(
+            (senderAccount.vaultCounter != 0) || (operators[_accountOwner][_sender]),
```
antoncoding, line 69: why do we need to check senderAccount.vaultCounter? What would happen if this is the first operation?

contracts/Controller.sol
```diff
@@ -124,38 +130,89 @@ contract Controller is Ownable {
...
+        OtokenInterface otoken = OtokenInterface(_otoken);
+
+        address otokenUnderlyingAsset = otoken.underlyingAsset();
+        address otokenStrikeAsset = otoken.strikeAsset();
+        address otokenCollateralAsset = otoken.collateralAsset();
+        uint256 otokenExpiryTimestamp = otoken.expiryTimestamp();
+
+        bytes32 batch = keccak256(
+            abi.encode(otokenUnderlyingAsset, otokenStrikeAsset, otokenCollateralAsset, otokenExpiryTimestamp)
+        );
```
antoncoding, line 171: use getBatch()

contracts/Controller.sol
```diff
@@ -124,38 +130,89 @@ contract Controller is Ownable {
...
+        return
+            keccak256(
+                abi.encode(
+                    otoken.underlyingAsset(),
+                    otoken.strikeAsset(),
+                    otoken.collateralAsset(),
+                    otoken.expiryTimestamp()
+                )
+            );
+    }
```
antoncoding, line 194: I think we should have this helper function in another contract like `OUtil.sol` , and both `Controller` and `Calculator` can inherit it.

contracts/Controller.sol
haythem96: yeah but still this function will need the `otoken.expiryTimestamp()`, so if I use `getBatch` will make to external call to `otoken`, so kept getBatch just for external use.

contracts/Controller.sol
haythem96: Where does calculator need it ?

contracts/Controller.sol
```diff
@@ -124,38 +130,89 @@ contract Controller is Ownable {
...
+            address calculatorModule = AddressBookInterface(addressBook).getMarginCalculator();
+            MarginCalculatorInterface calculator = MarginCalculatorInterface(calculatorModule);
+
+            (uint256 netValue, ) = calculator.getExcessMargin(vault, vault.shortOtokens[0]);
+            vault.collateralAmounts[0] = netValue;
+
+            return vault;
+        }
+
+        return vault;
```
antoncoding, line 150:
```suggestion

        if ((vault.shortOtokens.length == 0) || (!isExpired(vault.shortOtokens[0]))) 
            return vault;
        
        // if there's short and it's expired 
        address calculatorModule = AddressBookInterface(addressBook).getMarginCalculator();
        MarginCalculatorInterface calculator = MarginCalculatorInterface(calculatorModule);
       (uint256 netValue, ) = calculator.getExcessMargin(vault, vault.shortOtokens[0]);
        vault.collateralAmounts[0] = netValue;

       return vault;
        }        
```
Personally prefer guard clauses.

contracts/Controller.sol
antoncoding: But it looks so redundant. @aparnakr thoughts?

contracts/Controller.sol
haythem96: Any reason for that ?

contracts/Controller.sol
antoncoding: It needs to get the batch price too. (for getExpiredCashValue)

contracts/Controller.sol
haythem96: yeah but two external call to `otoken` will consume more gas.

antoncoding (dismissed):
🔥

contracts/Controller.sol
antoncoding: just a normal pattern to avoid using big if-else blocks.

antoncoding: Simple reason is that when people read til Line 150, they will already forget what the vault is for. If we just need to return it, better do it first, than deal with another logic

contracts/Controller.sol
haythem96: Will move it in another PR once Calculator get merged.

contracts/Controller.sol
haythem96: Updated this.

contracts/Controller.sol
```diff
@@ -47,25 +51,25 @@ contract Controller is Ownable {
...
     /**
-     * @notice modifier to check if sender is an authorized vault operator
+     * @notice modifier to check if sender is an account owner or an authorized account operator
      * @param _sender sender address
+     * @param _accountOwner account owner address
      */
-    // modifier isAuthorized(address _sender) {
-    //     _;
-    // }
+    modifier isAuthorized(address _sender, address _accountOwner) {
```
aparnakr, line 65: why not just have this be msg.sender? Does this need to take in _sender as an input?

contracts/Controller.sol
```diff
@@ -124,38 +128,88 @@ contract Controller is Ownable {
     /**
      * @notice Return a vault balances, depend of the short option expiry
```
aparnakr, line 129:
```suggestion
     * @notice Return a vault's balances, depend of the short option expiry
```

contracts/Controller.sol
```diff
@@ -124,38 +128,88 @@ contract Controller is Ownable {
...
      * @dev if vault have no short option or issued option is not expired yet, return vault, else get excess margin and return it as collateral amount inside Vault struct.
```
aparnakr, line 130:
```suggestion
     * @dev if vault has no short option or the issued option is not expired yet, return the vault, else call get excess margin and return it as collateral amount inside Vault struct.
```

contracts/Controller.sol
```diff
@@ -124,38 +128,88 @@ contract Controller is Ownable {
...
+     * @param _owner account owner.
      * @param _vaultId vault.
      * @return Vault struct
      */
-    // function getVaultBalances(address _owner, uint256 _vaultId) external view returns (MarginAccount.Vault memory) {
-    // }
+    function getVaultBalances(address _owner, uint256 _vaultId) external view returns (MarginAccount.Vault memory) {
+        MarginAccount.Vault memory vault = getVault(_owner, _vaultId);
+
+        // if there is no minted option or option not expired yet
```
aparnakr, line 138:
```suggestion
        // if there is no minted short option or the short option not expired yet
```

contracts/Controller.sol
```diff
@@ -124,38 +128,88 @@ contract Controller is Ownable {
...
+        if ((vault.shortOtokens.length == 0) || (!isExpired(vault.shortOtokens[0]))) return vault;
+
+        // if there's short and it's expired
```
aparnakr, line 141:
```suggestion
        // if there is a short option and it has expired
```

contracts/Controller.sol
```diff
@@ -124,38 +128,88 @@ contract Controller is Ownable {
...
+        address calculatorModule = AddressBookInterface(addressBook).getMarginCalculator();
+        MarginCalculatorInterface calculator = MarginCalculatorInterface(calculatorModule);
+
+        (uint256 netValue, ) = calculator.getExcessMargin(vault, vault.shortOtokens[0]);
```
aparnakr, line 145:
```suggestion
        (uint256 netValue, ) = calculator.getExcessMargin(vault, vault.shortOtokens[0].collateral);
```

aparnakr: GetExcessMargin only takes USDC or ETH as valid denominated assets right?

contracts/Controller.sol
```diff
@@ -124,38 +128,88 @@ contract Controller is Ownable {
...
+        vault.collateralAmounts[0] = netValue;
```
aparnakr, line 146: You should check if the vault has any collateral and return the min(vault.collateral, netValue)

aparnakr: https://docs.google.com/document/d/13xDJoPqfR6H3N_AMdBuSV8t20pEGiUqo4kCOJkBEV3o/edit?disco=AAAAKICQk3k see

contracts/Controller.sol
aparnakr: no strong thoughts on this one. whichever is easier to understand imo.

contracts/Controller.sol
```diff
@@ -124,38 +128,88 @@ contract Controller is Ownable {
...
+        return now > otokenExpiryTimestamp;
+    }
+
+    /**
+     * @notice Return a specific vault.
+     * @param _owner account owner.
+     * @param _vaultId vault.
+     * @return Vault struct
+     */
+    function getVault(address _owner, uint256 _vaultId) public view returns (MarginAccount.Vault memory) {
```
aparnakr, line 210: Does this function need to be public or is internal enough ?

contracts/Controller.sol
haythem96: nope, thought just for more clarity.

contracts/Controller.sol
haythem96: yeah fixed that in the later PR but forgot to do that here, will fix this.

contracts/Controller.sol
haythem96: Okay got that, but in the spec of `getVaultBalances()` it just say:
 ```
After Expiry:
CollateralAmounts = Controller.getExcessMargin(vault, vault.short.collateral);
Return vault {long, short, collateral: collateralAmount}
```
We should update that.

File: contracts/Controller.sol (hunk 12 above)
Comment (haythem96) [reply contracts/Controller.sol:210]: Why internal ? can be public to and used for any case IMO

Comment (aparnakr) [reply contracts/Controller.sol:212]: is it used publicly? Just don't want to confuse users since this won't give the accurate collateral balances

File: test/controller.test.ts [hunk 13]
```diff
@@ -30,9 +75,50 @@ contract('Controller', ([owner1, operator1, random]) => {
...
+
+      const expectedResutl = await oracle.isDisputePeriodOver(batch, otokenExpiryTimestamp)
+      assert.equal(await controller.isPriceFinalized(otoken.address), expectedResutl, 'Price is not finalized')
+    })
+  })
+
+  describe('Expiry', () => {
+    it('should check if otoken expired', async () => {
+      assert.equal(await controller.isExpired(otoken.address), false, 'Otoken expiry check mismatch')
     })
```
aparnakr, line 122: test a scenaro where the otoken has expired

test/controller.test.ts
```diff
@@ -30,9 +75,50 @@ contract('Controller', ([owner1, operator1, random]) => {
...
+      await oracle.setDisputePeriod(batchOracle.address, disputePeriod)
+      await oracle.setBatchUnderlyingPrice(batch, expiryTimestampMock, priceMock)
+    })
+
+    it('should check if price is finalized or not', async () => {
+      const batch = await controller.getBatch(otoken.address)
+      const otokenExpiryTimestamp = new BigNumber('1598374220')
+
+      const expectedResutl = await oracle.isDisputePeriodOver(batch, otokenExpiryTimestamp)
+      assert.equal(await controller.isPriceFinalized(otoken.address), expectedResutl, 'Price is not finalized')
```
aparnakr, line 115:
```suggestion
      assert.equal(await controller.isPriceFinalized(otoken.address), expectedResult, 'Price is not finalized')
```

test/controller.test.ts
```diff
@@ -30,9 +75,50 @@ contract('Controller', ([owner1, operator1, random]) => {
...
+
+      await oracle.setBatchOracle(batch, batchOracle.address)
+      await oracle.setDisputePeriod(batchOracle.address, disputePeriod)
+      await oracle.setBatchUnderlyingPrice(batch, expiryTimestampMock, priceMock)
+    })
+
+    it('should check if price is finalized or not', async () => {
+      const batch = await controller.getBatch(otoken.address)
+      const otokenExpiryTimestamp = new BigNumber('1598374220')
+
```
aparnakr, line 113: can you test it is not finalized before price is set and before dispute period?

test/controller.test.ts
```diff
@@ -30,9 +75,50 @@ contract('Controller', ([owner1, operator1, random]) => {
 
   describe('Account operator', () => {
     it('should set operator', async () => {
-      await controller.setOperator(operator1, true, {from: owner1})
+      await controller.setOperator(accountOperator1, true, {from: accountOwner1})
 
-      assert.equal(await controller.isOperator(owner1, operator1), true, 'Operator address mismatch')
+      assert.equal(await controller.isOperator(accountOwner1, accountOperator1), true, 'Operator address mismatch')
+    })
```
aparnakr, line 81: can you test that it is initially not the operator, and then is the operator? 

Also test that once it is the operator and the permissions are revoked, it is no longer the operator? 

Also test that only the account owner can set the operator? The operator cannot set themselves/ remove themselves as operator

contracts/Controller.sol
haythem96: Yeah I think we need it public to show vault data right ?

aparnakr (dismissed):
LGTM! Need some more unit tests, but other than that good to merge! :)

antoncoding:
You got a typo in your test file, so the CI test is not passing

contracts/Controller.sol
antoncoding: it's just one more external call to read expiryTimeStamp, and you save 3 lines of code, I think it's worth.

contracts/Controller.sol
antoncoding: We need this. So users can easily use getVault and than getExcessMargin on their own.

contracts/Controller.sol
```diff
@@ -158,38 +158,43 @@ contract Controller is Ownable {
...
-    function getBatch(address _otoken) external view returns (bytes32) {
+    function getBatchDetails(address _otoken)
+        public
+        view
+        returns (
+            bytes32,
+            address,
+            address,
+            address,
+            uint256
```
antoncoding, line 179: can you name these variables, otherwise it's hard to get any information from this function signature

haythem96: We usually don't name return veriables, let't keep it that way as the name are in the return statement IMO.

antoncoding (approved):
Can you add Controller.sol to be ignored by coverage test? It would be easier for us to keep track of the coverage report for other PRs if we can keep the master at 100%.

haythem96:
> Can you add Controller.sol to be ignored by coverage test? It would be easier for us to keep track of the coverage report for other PRs if we can keep the master at 100%.

@aparnakr thoughts ?

test/controller.test.ts
```diff
@@ -95,31 +107,80 @@ contract('Controller', ([owner, accountOwner1, accountOperator1, random]) => {
...
+      // increase time after locking period
+      await time.increase(61)
+      await oracle.setBatchUnderlyingPrice(batch, expiryTimestampMock, priceMock)
+      // increase time after dispute period
+      await time.increase(60)
 
-      const expectedResutl = await oracle.isDisputePeriodOver(batch, otokenExpiryTimestamp)
-      assert.equal(await controller.isPriceFinalized(otoken.address), expectedResult, 'Price is not finalized')
+      const expectedResutl = await oracle.isDisputePeriodOver(batch, expiryTimestampMock)
+      assert.equal(await controller.isPriceFinalized(otoken.address), expectedResutl, 'Price is not finalized')
```
aparnakr, line 160: can you also expectResult true or false? Is this testing for price is finalized or not finalized?

haythem96: This one is testing for price finalized:
- L155 price pushed
- L157 dispute period over

haythem96: @aparnakr updated `expectedResutl=true` so it doesn't make confusion.

haythem96: @aparnakr I updated all expected result to be a const input, idk why I assigned expected result from a call to the contract, def the wrong way to do it.

test/controller.test.ts
```diff
@@ -108,11 +108,24 @@ contract('Controller', ([owner, accountOwner1, accountOperator1, random]) => {
...
+        usdc.address,
+        usdc.address,
+        new BigNumber(200).times(new BigNumber(10).exponentiatedBy(18)),
+        new BigNumber(await time.latest()),
+        true,
+      )
+
+      const batch = (await controller.getBatchDetails(expiredOtoken.address))[0]
       // set batch oracle
       await oracle.setBatchOracle(batch, batchOracle.address)
```
aparnakr, line 130: maybe you shouldn't be able to setBatchOraclePrice if the locking and dispute period is set in the first place?

aparnakr: isn't that weird if you can set the price because then it automatically becomes finalized?

haythem96: mm what do u mean ?

haythem96:
> Being able to setBatchPrice before the disputePeriod and lockingPeriod have been set is a little weird imo. This is probably an issue at the oracle level though, not in the controller, so still needs to be resolved in a separate issue, but for now this can be merged

I don't understand this ? where do u see `setBatchUnderlyingPrice` before setting locking and dispute period ?

haythem96:
> > Being able to setBatchPrice before the disputePeriod and lockingPeriod have been set is a little weird imo. This is probably an issue at the oracle level though, not in the controller, so still needs to be resolved in a separate issue, but for now this can be merged
> 
> I don't understand this ? where do u see `setBatchUnderlyingPrice` before setting locking and dispute period ?

Also solidity initiate uint256 by default to zero, so if a batch oracle does not have a set locking/dispute period, then it is by default zero... but even in that case that didin't happen in the tests above!

test/controller.test.ts
aparnakr: oh whoops I read it as price. nvm. all good.

contracts/Controller.sol
aparnakr: wait why do we need this, still don't understand

aparnakr: when will users use the info about their vault? They really only care about how much actual collateral they have right, not colalteral that doesn't belong to them

aparnakr: this will confuse developers more imo

PR #92, opened 2020-08-25
Open vault functionality

haythem96:
# Task: Feature Name:

Controller module: open vault functionality

## High Level Description

Should be reviewed/merged after merging #89 .

### Code

- [x] Unit test 100% coverage

### Documentation


contracts/Controller.sol
```diff
@@ -91,8 +98,16 @@ contract Controller is Ownable {
...
-    // }
+    function operate(Actions.ActionArgs[] memory _actions) external isNotPaused nonReentrant {
+        MarginAccount.Vault memory vault = _runActions(_actions);
+
+        address calculatorModule = AddressBookInterface(addressBook).getMarginCalculator();
+        MarginCalculatorInterface calculator = MarginCalculatorInterface(calculatorModule);
+
+        if (vault.shortOtokens.length > 0) {
+            calculator.isValidState(vault, vault.shortOtokens[0]);
+        }
```
haythem96, line 109: Not sure about this, if we should only call `isValidState` when there is a short otoken in vault.
But can't pass an invalid index `vault.shortOtokens[0]` when  there is no short otoken.

contracts/libs/Actions.sol
```diff
@@ -3,11 +3,15 @@
...
 
+import {SafeMath} from "../packages/oz/SafeMath.sol";
+
 /**
  *
  */
 // solhint-disable-next-line no-empty-blocks
 library Actions {
+    using SafeMath for uint256;
+
```
aparnakr, line 14: what is this used for?

contracts/libs/MarginAccount.sol
```diff
@@ -10,12 +10,6 @@ import {SafeMath} from "../packages/oz/SafeMath.sol";
 library MarginAccount {
     using SafeMath for uint256;
 
-    // Account is a struct corresponding to a user that describes how many vaults that user has.
-    struct Account {
-        // the number of vaults a user has, starting at index 0 and incrementing. vaultCounter sorted chronologically.
-        uint256 vaultCounter;
-    }
-
```
aparnakr, line 18: Don't remove this! I need it for invariant testing lol. I need to know how many vault counters are there on chain.

contracts/libs/MarginAccount.sol
```diff
@@ -10,12 +10,6 @@ import {SafeMath} from "../packages/oz/SafeMath.sol";
 library MarginAccount {
     using SafeMath for uint256;
 
-    // Account is a struct corresponding to a user that describes how many vaults that user has.
```
aparnakr, line 13: Don't remove this

contracts/libs/MarginAccount.sol
antoncoding: We can just stored `mapping (address => uint256 )vaultCounter` in the controller.

contracts/libs/MarginAccount.sol
aparnakr: Actually nvm its good!

contracts/libs/MarginAccount.sol
aparnakr: nvm its good! I saw the other new implementation

contracts/Controller.sol
```diff
@@ -25,7 +31,8 @@ contract Controller is Ownable {
     address internal addressBook;
 
     /// @dev mapping between owner address and account structure
-    mapping(address => MarginAccount.Account) internal accounts;
+    //mapping(address => MarginAccount.Account) internal accounts;
```
aparnakr, line 34:
```suggestion
```

contracts/Controller.sol
```diff
@@ -122,57 +137,157 @@ contract Controller is Ownable {
...
+        address otokenUnderlyingAsset = otoken.underlyingAsset();
+        address otokenStrikeAsset = otoken.strikeAsset();
+        address otokenCollateralAsset = otoken.collateralAsset();
+        uint256 otokenExpiryTimestamp = otoken.expiryTimestamp();
+
+        bytes32 batch = keccak256(
+            abi.encode(otokenUnderlyingAsset, otokenStrikeAsset, otokenCollateralAsset, otokenExpiryTimestamp)
+        );
+
+        return oracle.isDisputePeriodOver(batch, otokenExpiryTimestamp);
```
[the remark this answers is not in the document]
antoncoding: Never mind, working on it now.

contracts/Controller.sol
```diff
@@ -122,57 +137,157 @@ contract Controller is Ownable {
...
-    // }
+    function _runActions(Actions.ActionArgs[] memory _actions) internal returns (MarginAccount.Vault memory) {
+        MarginAccount.Vault memory vault;
+
+        for (uint256 i = 0; i < _actions.length; i++) {
+            Actions.ActionArgs memory action = _actions[i];
+            Actions.ActionType actionType = action.actionType;
+
+            uint256 actionVaultId;
+            bool isActionVaultStored;
```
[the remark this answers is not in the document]
haythem96: I don't really like writing solidity in that way, as uint256 by default equal to 0 and bool by default equal to false, idk why crytic show those messages.

contracts/Controller.sol
```diff
@@ -122,57 +137,157 @@ contract Controller is Ownable {
...
+                    actionVaultId,
+                    action.vaultId,
+                    isActionVaultStored
+                );
+                _openVault(Actions._parseOpenVaultArgs(action));
+            }
+        }
+
+        return vault;
+    }
```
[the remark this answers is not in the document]
haythem96: yeah because in this case it is just the open vault action...
`_runActions` should return the vault to pass it to calculator for `isValideState`, even if there is not modified state, this function should return something

contracts/Controller.sol
haythem96: This doesn't sounds good for me, why the calculator need `address(0)` even if there is no short otoken ?

contracts/Controller.sol
```diff
@@ -122,57 +137,157 @@ contract Controller is Ownable {
...
+    function _checkActionVault(
+        uint256 _prevActionVaultId,
+        uint256 _currActionVaultId,
+        bool _isActionVaultStored
+    ) internal returns (uint256, bool) {
+        if (_isActionVaultStored) {
+            require(_prevActionVaultId == _currActionVaultId, "Controller: can not run actions on different vaults");
+        }
+        return (_currActionVaultId, true);
+    }
```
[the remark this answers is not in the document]
haythem96: The question is, do we want users/devs to pass in vaultId when they are creating new one ?
I have 2 vaults, should I query the vaultCounter and increment that and pass it in the action args when creating a new vault ?

contracts/Controller.sol
```diff
@@ -25,7 +31,7 @@ contract Controller is Ownable {
     address internal addressBook;
 
     /// @dev mapping between owner address and account structure
```
aparnakr, line 33: can you also add an array or some way to know on chain all the acocunt owners? (need it for invariant testing that all the collateral sums on all vaults <= total collateralAsset in margin pool)

haythem96: I think we don't need for a functionality we shouldn't add it! thoughts @antoncoding ?

contracts/Controller.sol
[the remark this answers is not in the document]
antoncoding: why can't the `operate` function get the vault again on its own after you run `_runActions`?

contracts/Controller.sol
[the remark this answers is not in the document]
antoncoding: I think this is a coding style issue, I personally think it's clearer to initialize with value, it also tell people the expected value at creation.
Crytic shows those message for a reason, I'm convinced it's better in terms of readability.

contracts/Controller.sol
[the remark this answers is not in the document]
antoncoding: for the openVault action they don't need to pass in, but for other operations they have to.
So they should have OpenVault(), depositCollateral({vaultId = 1})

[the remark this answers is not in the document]
antoncoding: or if it's easier for the validation, we can required the open Vault action to pass in vaultId as well

contracts/Controller.sol
antoncoding: @aparnakr If that's for fuzz testing, can't we only have that in the echidna contract? 
I think we shouldn't add them. Arrays on chain are pretty dangerous.

contracts/Controller.sol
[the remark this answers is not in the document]
haythem96: what's the benefits of that, if there is any ?

contracts/Controller.sol
antoncoding: yeah right now that parameter is really redundant, I feel like it would make everyone's life easier if we can remove it for now. Our original idea was to keep the interface consistent for future calculator updates, not sure if we still a requirement to hold.

contracts/Controller.sol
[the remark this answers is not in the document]
haythem96: I updated that, basically thought about it from dev perspective, so now a developer should query `getAccountVaultCounter` and increment that and pass it as `vaultId` when creating new vault and manipulating that. So based on that, vault id start from 1 and not 0.

contracts/Controller.sol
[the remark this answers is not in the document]
antoncoding: I think we should keep it consistent that don't return value for non-view functions.
The only function that calls runActions is `operate`, and it can get this variable on its own, so why do we need the runActions to return it?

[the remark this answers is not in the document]
haythem96: I don't remember the reason behind the return actually, following spec. But yeah I agree getting that at the end once better than assigning it at the end of every action and returning that.

contracts/Controller.sol
haythem96: Also `isValidState()` return bool right? so I should probably add a require statement there.

contracts/Controller.sol
aparnakr: hmm we would basically need to add to that array every time an account was created for echidna. What is the dangerous part of having it on chain?

aparnakr: we can have a different contract we are running echidna on, but every additional change to the contract which the echidna contract is running on is a little more out of date compared to the one we are deploying, so we ideally don't want to have too much different data

contracts/Controller.sol
```diff
@@ -91,8 +97,10 @@ contract Controller is Ownable {
      * @dev can only be called when system is not paused
      * @param _actions array of actions arguments
      */
-    // function operate(Actions.ActionArgs[] memory _actions) external isNotPaused {
-    // }
+    function operate(Actions.ActionArgs[] memory _actions) external isNotPaused nonReentrant {
+        _runActions(_actions);
+        _verifyFinalState(_actions[0].owner, _actions[0].vaultId);
```
aparnakr, line 102: The 0th action may now always have an owner and vaultId, example you could call exercise first and then call a bunch of other actions. You need to get the first action that has a vault owner and id from runActions and pass that into verifyFinalState

contracts/Controller.sol
```diff
@@ -122,57 +130,155 @@ contract Controller is Ownable {
...
+        for (uint256 i = 0; i < _actions.length; i++) {
+            Actions.ActionArgs memory action = _actions[i];
+            Actions.ActionType actionType = action.actionType;
+
+            uint256 prevActionVaultId;
+
+            if (actionType == Actions.ActionType.OpenVault) {
+                // check if this action is manipulating the same vault as all other actions, other than SettleVault
+                require(
+                    prevActionVaultId == 0 || action.vaultId == prevActionVaultId,
```
aparnakr, line 243: hmm what if the prevActionVaultId is 0? So if someone passes an action on vault 0 and an action on vault 1, won't this condition pass?

Maybe there should be a bool of isVaultIdAssigned?

contracts/Controller.sol
```diff
@@ -122,57 +130,155 @@ contract Controller is Ownable {
...
+
+    /**
+     * @notice verify vault final state after executing all actions
+     * @param _accountOwner account owner address
+     * @param _vaultId vault id related to the vault that will be checked
+     */
+    function _verifyFinalState(address _accountOwner, uint256 _vaultId) internal view {
+        MarginAccount.Vault memory vault = vaults[_accountOwner][_vaultId];
+
+        if (vault.shortOtokens.length > 0) {
```
aparnakr, line 261: shouldn't this check be in isValidState function? 

Also, why is this check there? Even if short.length is 0 we should check this right?

haythem96: if short.length is 0 then `vault.shortOtokens[0]` will revert with invalid opcode as we are accessing an invalid index

contracts/Controller.sol
haythem96: Yeah, so the old implementation make sense now
```
 function _checkActionVault(
        uint256 _prevActionVaultId,
        uint256 _currActionVaultId,
        bool _isActionVaultStored
    ) internal returns (uint256, bool) {
        if (_isActionVaultStored) {
            require(_prevActionVaultId == _currActionVaultId, "Controller: can not run actions on different vaults");
        }
        return (_currActionVaultId, true);
    }
```
@antoncoding thoughts ?

contracts/Controller.sol
haythem96: Great catch, so we definitely need to return that from the `_verifyFinalState`. // cc @antoncoding

haythem96: updated this.

contracts/Controller.sol
haythem96: updated this.

contracts/Controller.sol
```diff
@@ -230,42 +235,63 @@ contract Controller is ReentrancyGuard, Ownable {
...
+                    action.vaultId,
+                    isActionVaultStored
                 );
-                prevActionVaultId = action.vaultId;
 
                 _openVault(Actions._parseOpenVaultArgs(action));
             }
         }
+
+        return vault;
```
aparnakr: vault isn't initialized to anything though is it?

haythem96: because when user open a vault, it just increment vault counter in the mapping, doesn't modify any vault, in that case it will return an empty vault.

contracts/Controller.sol
```diff
@@ -230,42 +235,63 @@ contract Controller is ReentrancyGuard, Ownable {
...
-     * @param _accountOwner account owner address
-     * @param _vaultId vault id related to the vault that will be checked
+     * @param _vault final vault state
      */
-    function _verifyFinalState(address _accountOwner, uint256 _vaultId) internal view {
-        MarginAccount.Vault memory vault = vaults[_accountOwner][_vaultId];
-
-        if (vault.shortOtokens.length > 0) {
+    function _verifyFinalState(MarginAccount.Vault memory _vault) internal view {
+        if (_vault.shortOtokens.length > 0) {
```
aparnakr: is there an issue with verifying final state if no short tokens in a vault?

haythem96: yeah, `vault.shortOtokens[0]` will revert with invalid opcode error, because w ebasically accessing index 0 which is an invalid index.

Valid indexes in any array are between 0 and array.length - 1.

contracts/Controller.sol
aparnakr: ahh ok. for other actions it needs to be testsed differently then.

antoncoding (approved):

PR #94, opened 2020-08-27
Refactor Oracle structure.

antoncoding:
# Task: Improve Oracle

## High Level Description

* Refactor the Oracle contract into a 2 layer structure: **Oracle** + **Pricer**.
* The oracle contract no longer takes in `batch` as parameter, instead we query every data based on `asset` and `expiry`. 
* There is an internal mapping which maps  `asset` to `pricer`.
* We can implement arbitrary Pricer for each `asset`. In this PR we already include a `ChainLinkPricer`.

### Pricer

One Pricer can serve multiple assets, e.g. ChainLinkPricer can report price for ETH, WETH and more.

Only 1 function is required to be compliant with the `PricerInterface`:
* `getPrice(asset)`: return the LIVE price for that asset, scaled to 1e18.

But as a pricer, it also need to write price at expiry back to oracle. 
* the pricer has to call oracle.setExpiryPrice() to write back the asset's price at expiry to the oracle.
* eg. The `ChainLinkPricer` has a function `setExpiryPriceToOracle()` that write data back to the oracle.
 
#### Get Live Price
* getPrice(asset): return pricer.getPrice() for that asset

#### Get Expiry Price (for settlement)
* getExpiryPrice(asset, expiry)

### Checklist
- [x] 100% coverage

## Notes:
I updated the Controller tests, and remove lots of logic in MockOracle. I think it's easier to write tests with simple Mock contracts.


haythem96:
We should remove all batch usage in the Oracle and any Otoken usage and just map asset address by oracle address used.
For the locking and dispute period it has nothing to do also with Otoken, locking and dispute period is for the used oracle. (map asset to oracle address, map oracle to locking period, map oracle to dispute period).

haythem96:
> We should remove all batch usage in the Oracle and any Otoken usage and just map asset address by oracle address used. Also by mapping asset to oracle, we will not need to push price twice for option that have same underlying+expiry but different collateral.
> 
> For the locking and dispute period it has nothing to do also with Otoken, locking and dispute period is for the used oracle. (map asset to oracle address, map oracle to locking period, map oracle to dispute period).

So `getUnderlyingPriceAtExpiry` can just be `getPrice(address asset)` and no need for the `getBatchPrice()`

antoncoding:
> So `getUnderlyingPriceAtExpiry` can just be `getPrice(address asset)` and no need for the `getBatchPrice()`
But if we only have getPrice(asset), how can we get the price of underlying specifically at expiry?
IMO we either need getPrice(asset, timestamp), or `getUnderlyingPriceAtExpiry(otoken)`

antoncoding:
Our oracle have 2 functionalities:
1. For oTokens specific functionality, so this one will require `getUnderlyingPirceAtExpiry(otoken)`, which is required for cash settle exercise. 
2. For collateral price and strike asset price before expiry. This should be generalized to getPrice(asset)

I think I can remove the batch usage and only use a [underlying][ timestamp] => price mapping for the first one, but we can't really remove otoken usage, (we need either `getPrice(asset, timestamp)`, or getUnderlyingPriceAtExpiry(otoken))

haythem96:
the `Oracle.sol` miss the function that read price anytime right ? e.g ctoken exchange rate before expiry ?

haythem96:
> the `Oracle.sol` miss the function that read price anytime right ? e.g ctoken exchange rate before expiry ?

We should add a `getPrice` in Oracle.sol that will call the `getPrice` in the Pricer to get live price, wether it is a chainlink oracle or compound one. If I understand this implementation correctly, to get cToken exchange rate, we need to call the Pricer in this case, right ?

contracts/ChainlinkPricer.sol
```diff
@@ -0,0 +1,72 @@
...
+     * @notice get live price for a asset.
+     * @dev ovverides the getPrice function in OpynPricerInterface.
+     * @param _asset the address of the asset
+     */
+    function getPrice(address _asset) external override view returns (uint256) {
+        require(assetAggregator[_asset] != address(0), "ChainLinkPricer: aggregator for the asset not set.");
+        int256 answer = AggregatorInterface(assetAggregator[_asset]).latestAnswer();
+        require(answer > 0, "ChainLinkPricer: price is lower than 0");
+        // todo: scaled to 10e18 // 39010460929
+        return uint256(answer * 1e10);
```
haythem96, line 46: Is this right ?

contracts/ChainlinkPricer.sol
```diff
@@ -0,0 +1,72 @@
...
+
+    /// @notice the opyn oracle address
+    OracleInterface public oracle;
+
+    /// @notice the aggregator for an asset
+    mapping(address => address) public assetAggregator;
+
+    /// @notice the mapping of price
+
+    constructor(address _oracle) public {
```
haythem96, line 21: Will be better to add aggregator address as argument here, and use one pricer for one feed, easier to understand and less permissions as we will not need the onlyOwner in this case.

contracts/ChainlinkPricer.sol
```diff
@@ -0,0 +1,72 @@
...
+     * @param _asset asset address
+     * @param _aggregator chainlink aggregator address
+     */
+    function setAggregator(address _asset, address _aggregator) external onlyOwner {
+        require(_aggregator != address(0), "ChainLinkPricer: Cannot set 0 address as aggregator");
+        assetAggregator[_asset] = _aggregator;
+    }
+
+    /**
+     * @notice get live price for a asset.
```
aparnakr, line 37:
```suggestion
     * @notice get live price for an asset.
```

contracts/ChainlinkPricer.sol
```diff
@@ -0,0 +1,72 @@
...
+     * @dev ovverides the getPrice function in OpynPricerInterface.
```
aparnakr, line 38:
```suggestion
     * @dev overides the getPrice function in OpynPricerInterface.
```

contracts/ChainlinkPricer.sol
aparnakr: wait why 1e10?

antoncoding: because chainlink aggregator scaled in 1e8.

contracts/ChainlinkPricer.sol
antoncoding: Thats a good idea, will change it ;)

contracts/ChainlinkPricer.sol
antoncoding: @aparnakr I think it's time we install this https://github.com/marketplace/typo-ci 🤣

antoncoding:
> > the `Oracle.sol` miss the function that read price anytime right ? e.g ctoken exchange rate before expiry ?
> 
> We should add a `getPrice` in Oracle.sol that will call the `getPrice` in the Pricer to get live price, wether it is a chainlink oracle or compound one. If I understand this implementation correctly, to get cToken exchange rate, we need to call the Pricer in this case, right ?

Yeah Added. 
I think for CompoundPricer, it can return Chainlink.latest() * exchangeRate. (So we make sure everything is denominated in USD)

contracts/Controller.sol
```diff
@@ -161,43 +161,13 @@ contract Controller is ReentrancyGuard, Ownable {
...
-        return oracle.isDisputePeriodOver(batch, otokenExpiryTimestamp);
-    }
-
-    /**
-     * @notice get batch for a specific otoken address
-     * @dev batch is the hash of option underlying, strike, collateral assets and the expiry timestamp
-     * @param _otoken otoken address
-     * @return batch hash in bytes32, batch underlying asset, batch strike asset, batch collateral asset, batch expiry timestamp
-     */
-    function getBatchDetails(address _otoken)
```
haythem96, line 174: Why removed this function ?

haythem96 (approved):

contracts/Controller.sol
antoncoding: we don't need batch ID in the whole system anymore, so I don't see how would we need this function for

contracts/Oracle.sol
```diff
@@ -68,183 +68,165 @@ contract Oracle is Ownable {
...
+     * @param _pricer pricer address
      * @param _lockingPeriod locking period
      */
-    function setLockingPeriod(address _oracle, uint256 _lockingPeriod) external onlyOwner {
-        oracleLockingPeriod[_oracle] = _lockingPeriod;
+    function setLockingPeriod(address _pricer, uint256 _lockingPeriod) external onlyOwner {
+        pricerLockingPeriod[_pricer] = _lockingPeriod;
 
-        emit OracleLockingPeriodUpdated(_oracle, _lockingPeriod);
+        emit PricerLockingPeriodUpdated(_pricer, _lockingPeriod);
```
aparnakr, line 177: should probably not allow locking period to be reset during locking period. Admin doesn't need that much power do they? Also it seems like a malicious admin could then keep locking up people's funds.

aparnakr: Same with dispute period, should not allow reset of the dispute period during a dispute period.

aparnakr: or should have some max caps on these?

test/controller.test.ts
```diff
@@ -261,31 +253,19 @@ contract('Controller', ([owner, accountOwner1, accountOperator1, random]) => {
...
-      await oracle.setLockingPeriod(batchOracle.address, lockingPeriod)
-      await oracle.setDisputePeriod(batchOracle.address, disputePeriod)
-    })
-
-    it('should return false when price is not pushed to Oracle yet', async () => {
-      assert.equal(
-        await controller.isPriceFinalized(expiredOtoken.address),
-        false,
-        'Price is not finalized because it is not stored yet',
-      )
```
aparnakr, line 277: can you still keep this test on should be false when price has not been finalized?

test/chainlinkPricer.test.ts
```diff
@@ -0,0 +1,138 @@
...
+    // aggregator have price in 1e8
+    const ethPrice = toChainLinkPrice(300)
+    before('mock data in weth aggregator', async () => {
+      await wethAggregator.setLatestAnswer(ethPrice)
+    })
+    it('should return the price in 1e18', async () => {
+      const price = await pricer.getPrice()
+      const expectedResult = new BigNumber(300).times(1e18)
+      assert.equal(price.toString(), expectedResult.toString())
+    })
```
aparnakr, line 70: change the price and test if it returns the new price in getPrice?

contracts/Oracle.sol
antoncoding: Do not allow setting dispute period during dispute period sounds good to me.

test/controller.test.ts
antoncoding: There were originally 2 tests for this:
1. before pushing the price to oracle, this should return false
2. after pushing the price to oracle but during dispute period, this should return false.

But if you look at the `setIsFinalized()` code, it gets the answer from  `oracle.getExpiryPrice(underlying, expiry)` which returns true or false. So we were basically testing:
1.  before pushing the price to oracle, `oracle.getExpiryPrice` should return false
2. after pushing the price to oracle but during dispute period, `oracle.getExpiryPrice` should return false.

Which are not unit tests for controller at all. 
Instead, we just need one test for this: When `oracle.getExpiryPrice` returns false, `isPriceFinalized` should return false. 

Now I simplify the test because I changed the MockOracle contract to make it simpler to mock returned data. (don't need to push price to Oracle, wait for dispute period... etc, just mock the answer of`getExpiryPrice` )

PR #96, opened 2020-08-27
Deposit long Otoken

haythem96:
# Task: Feature Name:
Controller module: Deposit long Otoken

## High Level Description

This PR add the functionality of depositing a long Otoken into a vault.
Should be reviewed/merged after #92 .

### Code

- [ ] Unit test 100% coverage

### Documentation


antoncoding:
One un-related feedback:
We should use rebase more... the git history is getting messy because it contain all the changes from the controller skeleton PR.

contracts/Controller.sol
```diff
@@ -310,7 +322,30 @@ contract Controller is ReentrancyGuard, Ownable {
      * @notice deposit long option into vault
      * @param _args DepositArgs structure
      */
-    // function _depositLong(Actions.DepositArgs memory _args) internal {}
+    function _depositLong(Actions.DepositArgs memory _args) internal returns (MarginAccount.Vault memory) {
+        require(_args.from == msg.sender, "Controller: depositor address and msg.sender address mismatch");
+
+        address whitelistModule = AddressBookInterface(addressBook).getWhitelist();
+        WhitelistInterface whitelist = WhitelistInterface(whitelistModule);
```
antoncoding, line 329: You can probably have an internal function called `getWhitelist()` that returns a WhitelistInterface, also the same for MarginPoolInterface, I can see we will use this a lot in the Controller. 

Will only save you 1 line of code but I think it helps readability a bit,  the naming of `whitelistModule` and `whitelist` is not that helpful 😂

haythem96: Yeah I agree, I'm gonna refactor this to something more fancy than that in another PR 😋

contracts/Controller.sol
```diff
@@ -310,7 +322,30 @@ contract Controller is ReentrancyGuard, Ownable {
      * @notice deposit long option into vault
      * @param _args DepositArgs structure
      */
-    // function _depositLong(Actions.DepositArgs memory _args) internal {}
+    function _depositLong(Actions.DepositArgs memory _args) internal returns (MarginAccount.Vault memory) {
+        require(_args.from == msg.sender, "Controller: depositor address and msg.sender address mismatch");
```
antoncoding, line 326: wasn't the conclusion for args.from also be acceptable to be the vault owner?
Let's say msg.sender is an operator, the operator can also specify the `owner` to be the args.from. In that case our marginPool will just try to pull assets from the vault owner.

antoncoding (approved):
Look good! The tests are pretty clean especially the part for multiple actions.
I think the args.sender discussion worth a whole other thread, no need to resolve for this PR 🔥

contracts/Controller.sol
aparnakr: for later down the line, not right now.

aparnakr: not in this version, we said its too dangerous to open that up.

PR #101, opened 2020-08-29
Withdraw long otoken

haythem96:
# Task: Feature Name:

Controller module: withdraw long otoken

## High Level Description

This PR add the needed code for the withdraw long otoken functionality

### Code

- [ ] Unit test 100% coverage

### Documentation


contracts/Controller.sol
```diff
@@ -75,6 +75,17 @@ contract Controller is ReentrancyGuard, Ownable {
         _;
     }
 
+    /**
+     * @notice modifier to check the validity of a specific vault id
+     * @param _accountOwner account owner address
+     * @param _vaultId vault id
+     */
+    modifier runPreprocessing(address _accountOwner, uint256 _vaultId) {
```
antoncoding, line 83: I like the idea of using modifier here, but the name `runPreprocessing ` sounds like an internal function instead of a modifier.
Maybe renane to: `onlyOnValidVault`?

``

contracts/mocks/MockMarginCalculator.sol
```diff
@@ -14,6 +15,61 @@ contract MockMarginCalculator {
...
+    function _getBatchId(OtokenInterface _otoken) internal view returns (bytes32 id) {
+        id = keccak256(
+            abi.encodePacked(
+                _otoken.underlyingAsset(),
+                _otoken.strikeAsset(),
+                _otoken.collateralAsset(),
+                _otoken.expiryTimestamp()
+            )
+        );
+    }
```
antoncoding, line 74: I think it's better not to use batchId for comparison, it's a waste of gas, we only need to compare they have to check they have 4 same variables, don't have to hash anything.

haythem96: Yeah I agree, we should change that in Calculator also.

test/controller.test.ts
```diff
@@ -530,6 +501,367 @@ contract('Controller', ([owner, accountOwner1, accountOperator1, random]) => {
...
+          },
+        ]
+
+        await secondLongOtoken.approve(marginPool.address, collateralToDeposit, {from: accountOwner1})
+        await expectRevert(
+          controller.operate(actionArgs, {from: accountOwner1}),
+          'MarginCalculator: Too many long otokens in the vault.',
+        )
+      })*/
+    })
```
antoncoding, line 542: Why removing this?

haythem96: Didn't remove it, that commented because related to the arrays length checks issue, will uncomment when I implement checks in Controller.

contracts/Controller.sol
```diff
@@ -322,7 +340,25 @@ contract Controller is ReentrancyGuard, Ownable {
...
+        require(now <= otoken.expiryTimestamp(), "Controller: can not withdraw an expired otoken");
+
+        vaults[_args.owner][_args.vaultId]._removeLong(address(otoken), _args.amount, _args.index);
+
+        address marginPoolModule = AddressBookInterface(addressBook).getMarginPool();
+        MarginPoolInterface marginPool = MarginPoolInterface(marginPoolModule);
+
+        marginPool.transferToUser(address(otoken), _args.to, _args.amount);
+
+        return vaults[_args.owner][_args.vaultId];
```
antoncoding, line 360: I don't think we need to return this. Same for `_depositLong `, if there're multiple actions, we will just waste the gas of updating this `vault memory`.
I think we should just let `_verifyFinalState ` function access the owner's vault directly, and that pass that to the calculator.

contracts/mocks/MockMarginCalculator.sol
antoncoding, line 74: Move to controller
Remove from mock

contracts/mocks/MockMarginCalculator.sol
```diff
@@ -14,6 +15,61 @@ contract MockMarginCalculator {
...
+        // ensure the long asset is valid for the short asset.
+        _checkIsMarginableLong(_vault);
+
         return true;
     }
+
+    function _checkIsValidSpread(MarginAccount.Vault memory _vault) internal pure {
+        require(_vault.shortOtokens.length <= 1, "MarginCalculator: Too many short otokens in the vault.");
+        require(_vault.longOtokens.length <= 1, "MarginCalculator: Too many long otokens in the vault.");
+        require(_vault.collateralAssets.length <= 1, "MarginCalculator: Too many collateral assets in the vault.");
```
antoncoding, line 41: We should move the checks to controller like you suggested. In the mock contract just use mocked return value without requirement clauses.

contracts/mocks/MockMarginCalculator.sol
```diff
@@ -14,6 +15,61 @@ contract MockMarginCalculator {
...
+        public
+        view
+        returns (bool isExcess)
+    {
+        // ensure the number of collateral, long and short array is valid.
+        _checkIsValidSpread(_vault);
+        // ensure the long asset is valid for the short asset.
+        _checkIsMarginableLong(_vault);
+
         return true;
```
antoncoding, line 35: Mock this value instead of always using true

antoncoding (changes requested):
Moving the checks to controller sound like a good idea to me. Will re-review after that's done

contracts/mocks/MockMarginCalculator.sol
haythem96: yeah, for now will keep it as it is as I'm mocking a true result, will update it in the mint otoken PR, in either way this function will always return a predefined result(will not external call to oracle...etc)

contracts/mocks/MockMarginCalculator.sol
haythem96: Will keep `_checkIsMarginableLong` in calculator, will move only function that checks storage state after MarginAccount lib functions.

contracts/mocks/MockMarginCalculator.sol
antoncoding: I'm gonna resolve this given we decide to keep the checks in calculator

test/controller.test.ts
antoncoding: Now you can remove all this ;)

contracts/Controller.sol
haythem96: Some actions have no account owner/vault id arguments, so it will require checking that, and based on action assign vault owner, vault id and call the `_verifyFinalState`... will think more about it and refactor it in another PR.

antoncoding: okay I got your point.

contracts/mocks/MockMarginCalculator.sol
```diff
@@ -5,15 +5,73 @@ pragma solidity =0.6.10;
...
+            "MarginCalculator: Short asset and amount mismatch."
+        );
+        require(
+            _vault.longOtokens.length == _vault.longAmounts.length,
+            "MarginCalculator: Long asset and amount mismatch."
+        );
+        require(
+            _vault.collateralAssets.length == _vault.collateralAmounts.length,
+            "MarginCalculator: Collateral asset and amount mismatch."
+        );
```
antoncoding, line 50: You should remove these logics from the mock contracts and just mock the returned values

antoncoding (approved):
🔥 🔥

contracts/mocks/MockMarginCalculator.sol
haythem96: There is no returned value from those functions, If I remove them from the mock, some use cases that should revert will not.

antoncoding: Updated: These checks are required for easier validation checks in the Controller unit tests, will keep it here

contracts/Controller.sol
```diff
@@ -147,9 +158,8 @@ contract Controller is ReentrancyGuard, Ownable {
         // if there is a short option and it has expired
         address calculatorModule = AddressBookInterface(addressBook).getMarginCalculator();
         MarginCalculatorInterface calculator = MarginCalculatorInterface(calculatorModule);
-        OtokenInterface otoken = OtokenInterface(vault.shortOtokens[0]);
 
-        (uint256 netValue, ) = calculator.getExcessMargin(vault, otoken.collateralAsset());
+        (uint256 netValue, ) = calculator.getExcessMargin(vault);
```
aparnakr, line 162: this decision needs to be discussed before we change it

contracts/Controller.sol
```diff
@@ -292,7 +302,11 @@ contract Controller is ReentrancyGuard, Ownable {
      * @notice deposit long option into vault
      * @param _args DepositArgs structure
      */
-    function _depositLong(Actions.DepositArgs memory _args) internal returns (MarginAccount.Vault memory) {
+    function _depositLong(Actions.DepositArgs memory _args)
+        internal
+        isValidVaultId(_args.owner, _args.vaultId)
```
aparnakr, line 307: Can you make this a require and not a modifier? Little harder to read

haythem96: u mean the `isValidVaultId` ? I think it will be used in many functions, better to keep it a modifier ?!

aparnakr: yeah, well its a little confusing to read that + is authorized and all as modifiers. Its easier to read them all if they are in the function.

aparnakr: When there are multiple checks, it becomes confusing to keep track of what has been done and not if we use too many modifiers.

aparnakr: it can still be a function, but doing it as a modifier makes it confusing

test/controller.test.ts
```diff
@@ -176,35 +176,6 @@ contract('Controller', ([owner, accountOwner1, accountOperator1, random]) => {
       )
     })
 
-    it('should revert opening multiple vaults in the same operate call', async () => {
-      const actionArgs = [
```
aparnakr, line 180: why remove this test?

aparnakr: This should still be true right?

contracts/Controller.sol
haythem96: @antoncoding thoughts ?

test/controller.test.ts
haythem96: yeah, removed by mistake, will add that.

haythem96: Added back.

test/controller.test.ts
```diff
@@ -530,6 +501,367 @@ contract('Controller', ([owner, accountOwner1, accountOperator1, random]) => {
...
+          collateralToWithdraw.toString(),
+          'Receiver long otoken balance mismatch',
+        )
+        assert.equal(vaultAfter.longOtokens.length, 1, 'Vault long otoken array length mismatch')
+        assert.equal(vaultAfter.longOtokens[0], ZERO_ADDR, 'Vault long otoken address after clearing mismatch')
+        assert.equal(
+          new BigNumber(vaultBefore.longAmounts[0]).minus(new BigNumber(vaultAfter.longAmounts[0])).toString(),
+          collateralToWithdraw.toString(),
+          'Long otoken amount in vault after withdraw mismatch',
+        )
```
aparnakr, line 814: add test to prevent withdrawing 0 options

test/controller.test.ts
```diff
@@ -530,6 +531,367 @@ contract('Controller', ([owner, accountOwner1, accountOperator1, random]) => {
...
+
+        await secondLongOtoken.approve(marginPool.address, collateralToDeposit, {from: accountOwner1})
+        await expectRevert(
+          controller.operate(actionArgs, {from: accountOwner1}),
+          'MarginCalculator: Too many long otokens in the vault.',
+        )
+      })
+    })
+
+    describe('withdraw long otoken', () => {
```
aparnakr, line 574: add a test to test "revert if you try to remove more long than you own"

haythem96: I removed that from testing document, that's basically testing the `SafeMath` lib right ? in this case `_vault.shortAmounts[_index] = _vault.shortAmounts[_index].sub(_amount)` in MarginAccount will revert.

aparnakr: yes, but still worth having that test, just to double check.

antoncoding: I like this test too

contracts/Controller.sol
antoncoding: I think internal function is easier to read, I feel like people usually use modifier for access control (to see who can call this function or not). I put other checks in internal functions

haythem96: updated this.

test/controller.test.ts
haythem96: added.

test/controller.test.ts
haythem96: added.

haythem96:
Made the suggested changes, will merge this one.

PR #105, opened 2020-08-31
Deposit collateral asset

haythem96:
# Task: Feature Name:

Controller: deposit collateral

## High Level Description

This PR add the needed code for the deposit collateral functionality.
Should be reviewed/merged after #101 .
### Code

- [ ] Unit test 100% coverage

### Documentation


contracts/Controller.sol
```diff
@@ -322,13 +376,58 @@ contract Controller is ReentrancyGuard, Ownable {
...
+            whitelist.isWhitelistedCollateral(_args.asset),
+            "Controller: asset is not whitelisted to be used as collateral"
+        );
+
+        vaults[_args.owner][_args.vaultId]._addCollateral(_args.asset, _args.amount, _args.index);
+
+        address marginPoolModule = AddressBookInterface(addressBook).getMarginPool();
+        MarginPoolInterface marginPool = MarginPoolInterface(marginPoolModule);
+
+        marginPool.transferToPool(_args.asset, _args.from, _args.amount);
```
aparnakr, line 425: might need to have a check on args.from == msg.sender depending on what decision we make on if its okay to allow vault operators to take money from the users' addresses directly

test/controller.test.ts
```diff
@@ -530,6 +536,645 @@ contract('Controller', ([owner, accountOwner1, accountOperator1, random]) => {
           'Controller: otoken used as collateral is already expired',
         )
       })
+
+      it('should revert when vault have more than 1 long otoken', async () => {
+        const expiryTime = new BigNumber(60 * 60) // after 1 hour
+        const collateralToDeposit = new BigNumber('20')
```
aparnakr, line 542:
```suggestion
        const longToDeposit = new BigNumber('20')
```

test/controller.test.ts
```diff
@@ -530,6 +536,645 @@ contract('Controller', ([owner, accountOwner1, accountOperator1, random]) => {
...
+    it('should revert depositing a collateral asset from a msg.sender different than arg.from', async () => {
+      const vaultCounter = new BigNumber(await controller.getAccountVaultCounter(accountOwner1))
+      assert.isAbove(vaultCounter.toNumber(), 0, 'Account owner have no vault')
+
+      const collateralToDeposit = new BigNumber('10')
+      const actionArgs = [
+        {
+          actionType: ActionType.DepositCollateral,
+          owner: accountOwner1,
+          sender: random,
```
aparnakr, line 1020: we should test this specifically for the vault operator as the sender and the user as the sender as well, depending on what decision we make there.

test/controller.test.ts
```diff
@@ -530,6 +536,645 @@ contract('Controller', ([owner, accountOwner1, accountOperator1, random]) => {
...
+      ]
+
+      await usdc.approve(marginPool.address, collateralToDeposit, {from: accountOwner1})
+      await expectRevert(
+        controller.operate(actionArgs, {from: accountOwner1}),
+        'MarginAccount: invalid collateral amount',
+      )
+    })
+
+    it('should revert when vault have more than 1 long otoken', async () => {
```
aparnakr, line 1061:
```suggestion
    it('should revert when vault have more than 1 collateral type', async () => {
```

test/controller.test.ts
haythem96: random can be any address, right ?

test/controller.test.ts
```diff
@@ -95,6 +95,11 @@ contract('Controller', ([owner, accountOwner1, accountOperator1, random]) => {
     await addressBook.setController(controller.address, {from: owner})
 
     assert.equal(await controller.systemPaused(), false, 'System is paused')
+
+    // make everyone rich
+    await usdc.mint(accountOwner1, new BigNumber('1000'))
+    await usdc.mint(accountOperator1, new BigNumber('1000'))
+    await usdc.mint(random, new BigNumber('1000'))
```
aparnakr, line 102: 🤣  🤑  💵

test/controller.test.ts
aparnakr: yes, but the reason i say we should test for the vault operator and vault owner is that if we decide that vault operator should be able to take money from vault owner, then that is something we need to ensure we test. If not, we need to test that its not possible.

antoncoding (approved):

test/controller.test.ts
haythem96: This is collateral asset not long otoken

aparnakr: hmm but you are calling the DepositLong operate action?

PR #107, opened 2020-08-31
Upgradeability

haythem96:
# Task: Upgradeability:

Upgradeability: proxy pattern implementation

## High Level Description

Upradeability implementation.

### Code

- [x] Unit test 100% coverage

### Documentation


haythem96:
Will keep this as the last PR

contracts/Controller.sol
```diff
@@ -40,11 +41,13 @@ contract Controller is ReentrancyGuard, Ownable {
...
      */
-    constructor(address _addressBook) public {
-        require(_addressBook != address(0), "Controller: Invalid address book");
+    function initialize(address _addressBook, address _owner) public initializer {
+        __Context_init_unchained();
+        __Ownable_init_unchained(_owner);
+        __ReentrancyGuard_init_unchained();
 
         addressBook = _addressBook;
     }
```
antoncoding, line 53:
```suggestion
    function initialize(address _addressBook) public initializer {
        __Context_init_unchained();
        __Ownable_init_unchained(msg.sender);
        __ReentrancyGuard_init_unchained();

        addressBook = _addressBook;
    }
```

haythem96: In this case the owner will not be the msg.sender.
The `AddressBook` will call this function, so the addressbook contract should be be the owner, the owner will be the addressbook contract's owner, which is a multisig or a DAO in a later stage.

https://github.com/opynfinance/GammaProtocol/blob/3248cffad4aa9d8bc87756fe9bd9b0bedb20c15f/contracts/AddressBook.sol#L223

antoncoding: I think only using 1 address make it more generalized.

haythem96: We can't do that as there is no function that send custom transactions in the AddressBook, better to keep it this way to separate permission.

antoncoding (approved):
I think we should rename `setController` to `updateController`

aparnakr (approved):

PR #109, opened 2020-09-02
Withdraw collateral asset from vault

haythem96:
# Task: Feature Name:

Controller: withdraw collateral asset

## High Level Description

This PR add the needed code for the withdraw collateral functionality.
Should be reviewed/merged after #105 .

### Code

- [ ] Unit test 100% coverage

### Documentation


contracts/Controller.sol
```diff
@@ -322,20 +364,90 @@ contract Controller is ReentrancyGuard, Ownable {
...
 
     /**
      * @notice withdraw collateral asset from vault
      * @dev only account owner or operator can withdraw long option from vault
      * @param _args WithdrawArgs structure
      */
-    // function _withdrawCollateral(Actions.WithdrawArgs memory _args) internal isAuthorized(_args.owner) {}
+    function _withdrawCollateral(Actions.WithdrawArgs memory _args)
+        internal
+        isAuthorized(msg.sender, _args.owner)
```
aparnakr, line 425: I would move this modifier also to be a function just for cleaner readability. @antoncoding thoughts?

haythem96: Should we change all modifiers in code base to functions ?

aparnakr (approved):
Looks great!! Can we also add a test for cannot remove more collateral than the vault owner has? After that good to merge!

haythem96:
> Looks great!! Can we also add a test for cannot remove more collateral than the vault owner has? After that good to merge!

Added.

contracts/Controller.sol
antoncoding: Yeah I think we can remove these kind of modifier usage for now (which involve `_arg.`)

contracts/Controller.sol
```diff
@@ -418,7 +420,34 @@ contract Controller is ReentrancyGuard, Ownable {
...
+        returns (MarginAccount.Vault memory)
+    {
+        require(checkVaultId(_args.owner, _args.vaultId), "Controller: invalid vault id");
+
+        MarginAccount.Vault memory vault = vaults[_args.owner][_args.vaultId];
+        if ((vault.shortOtokens.length > 0) && (vault.shortOtokens[0] != address(0))) {
```
antoncoding, line 431: better have an internal function for this logic, maybe called `_notEmpty`
`arr.length > 0 && arr[0] !=address(0) `

antoncoding (approved):
Looks good, left 1 suggestion that would help readability

PR #110, opened 2020-09-02
Compound pricer

antoncoding:
#  Compound Pricer

## High Level Description

Add compound pricer to get price of cToken. (real cToken value in term of USD)

A compound pricer will also need to use another price to get price of underlying.
For example, the cUSDC pricer will need a USDC pricer to work, and cETH pricer will need a ETH pricer to work.

### Code

- [x] Unit test 100% coverage

### Documentation

### DIFF
--- a/contracts/pricers/CompoundPricer.sol
+++ b/contracts/pricers/CompoundPricer.sol
@@ -1,12 +1,11 @@
 // SPDX-License-Identifier: UNLICENSED
 pragma solidity 0.6.10;
 
-import {AggregatorInterface} from "./interfaces/AggregatorInterface.sol";
-import {OracleInterface} from "./interfaces/OracleInterface.sol";
-import {OpynPricerInterface} from "./interfaces/OpynPricerInterface.sol";
-import {CTokenInterface} from "./interfaces/CTokenInterface.sol";
-import {ERC20Interface} from "./interfaces/ERC20Interface.sol";
-import {SafeMath} from "./packages/oz/SafeMath.sol";
+import {OracleInterface} from "../interfaces/OracleInterface.sol";
+import {OpynPricerInterface} from "../interfaces/OpynPricerInterface.sol";
+import {CTokenInterface} from "../interfaces/CTokenInterface.sol";
+import {ERC20Interface} from "../interfaces/ERC20Interface.sol";
+import {SafeMath} from "../packages/oz/SafeMath.sol";
 
 /**
  * @notice A Pricer contract for Compound Asset.
--- a/test/compoundPricer.test.ts
+++ b/test/compoundPricer.test.ts
@@ -44,7 +44,7 @@ contract('CompoundPricer', ([owner, random]) => {
   before('Deployment', async () => {
     // deploy mock contracts
     oracle = await MockOracle.new({from: owner})
-    weth = await MockERC20.new('WETH', 'WETH')
+    weth = await MockERC20.new('WETH', 'WETH', 18)
     cETH = await MockCToken.new('CETH', 'CETH')
     //
     chainlinkPricer = await MockPricer.new(weth.address, oracle.address)
--- a/test/compoundPricer.test.ts
+++ b/test/compoundPricer.test.ts
@@ -1,0 +1,134 @@
+import {
+  MockPricerInstance,
+  MockOracleInstance,
+  MockERC20Instance,
+  MockCTokenInstance,
+  CompoundPricerInstance,
+} from '../build/types/truffle-types'
+import BigNumber from 'bignumber.js'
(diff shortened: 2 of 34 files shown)


haythem96:
In this implementation, if we wanna push cUSDC price, we need to keep pushing USDC price also using a pricer?! It will be better to modify this and use `USDC price = 1` because it is always equal to 1. We can pricer other Pricer for different cToken when we need to.

antoncoding:
> In this implementation, if we wanna push cUSDC price, we need to keep pushing USDC price also using a pricer?! It will be better to modify this and use `USDC price = 1` because it is always equal to 1. We can pricer other Pricer for different cToken when we need to.

Yeah, or we can deploy a Pricer for USDC that always return `1e18`. I think that will be more flexible.

haythem96:
> > In this implementation, if we wanna push cUSDC price, we need to keep pushing USDC price also using a pricer?! It will be better to modify this and use `USDC price = 1` because it is always equal to 1. We can pricer other Pricer for different cToken when we need to.
> 
> Yeah, or we can deploy a Pricer for USDC that always return `1e18`. I think that will be more flexible.

I don't think we need that, we can just use 1e18 as it is always 1e18.

antoncoding:
> > > In this implementation, if we wanna push cUSDC price, we need to keep pushing USDC price also using a pricer?! It will be better to modify this and use `USDC price = 1` because it is always equal to 1. We can pricer other Pricer for different cToken when we need to.
> > 
> > 
> > Yeah, or we can deploy a Pricer for USDC that always return `1e18`. I think that will be more flexible.
> 
> I don't think we need that, we can just use 1e18 as it is always 1e18.

So we should set a `USDC` address in the oracle, and always return 1e18 for that? 

I also need to change the CompoundPricer to get underlying price from oracle, instead of a pricer, so we can just return 1e18 for USDC in the oracle instead of using a Pricer.

haythem96:
> > > > In this implementation, if we wanna push cUSDC price, we need to keep pushing USDC price also using a pricer?! It will be better to modify this and use `USDC price = 1` because it is always equal to 1. We can pricer other Pricer for different cToken when we need to.
> > > 
> > > 
> > > Yeah, or we can deploy a Pricer for USDC that always return `1e18`. I think that will be more flexible.
> > 
> > 
> > I don't think we need that, we can just use 1e18 as it is always 1e18.
> 
> So we should set a `USDC` address in the oracle, and always return 1e18 for that?
> 
> I also need to change the CompoundPricer to get underlying price from oracle, instead of a pricer, so we can just return 1e18 for USDC in the oracle instead of using a Pricer.

Oh I get ur point, so ur saying having a pricer for USDC is more generic than just adding an exception in the Oracle module for that, yeah sounds good to me then we can have an USDC pricer, but it should always return 1e18 without the need for a bot to push to it.

haythem96 (approved):
LGTM.

test/compoundPricer.test.ts
```diff
@@ -0,0 +1,133 @@
...
+      const newPrice = createScaledNumber(500)
+      await chainlinkPricer.setPrice(newPrice)
+      const cTokenPrice = await compoundPricer.getPrice()
+      const expectedResult = await underlyingPriceToCtokenPrice(new BigNumber(newPrice), exchangeRate, weth)
+      assert.equal(cTokenPrice.toString(), expectedResult.toString())
+    })
+    it('should revert if price is lower than 0', async () => {
+      await chainlinkPricer.setPrice('0')
+      await expectRevert(compoundPricer.getPrice(), 'CompoundPricer: underlying price is 0')
+    })
```
aparnakr, line 109: Can you add a test to see if it works for cUSDC, cDai as well? If cUSDC works, then it tests for the different decimals properly.

contracts/pricers/CompoundPricer.sol
```diff
@@ -0,0 +1,85 @@
...
+        oracle.setExpiryPrice(address(cToken), _expiryTimestamp, cTokenPrice);
+    }
+
+    /**
+     * @dev convert underlying price to cToken price.
+     * @param _underlyingPrice price of 1 underlying token (1e6 USDC, 1e18 WETH) in USD, scled by 1e18
+     * @return net worth of 1e8 cToken in USD, scaled by 1e18.
+     */
+    function _underlyingPriceToCtokenPrice(uint256 _underlyingPrice) internal view returns (uint256) {
+        uint256 underlyingDecimals = underlying.decimals();
```
aparnakr, line 80: this should always be a 1e18 number right?

aparnakr (approved):
Looks good to me, also add tests for cUSDC to check that it works for the different decimals.

PR #112, opened 2020-09-03
Whitelist module: blacklist functionalities

haythem96:
# Task: Feature Name:

Whitelist module: blacklist functionalities

## High Level Description

This PR change the following in the Whitelist module:
- Add blacklisting collateral, product and Otoken from owner.
- Refactored functions to allow for usage in both cases, whitelisting and blacklisting
- Remove require statement to check that owner not whitelisting an already whitelisted asset, less gas consumption for checks that does not harm the protocol.

### Code

- [x] Unit test 100% coverage

### Documentation

### DIFF
--- a/contracts/Whitelist.sol
+++ b/contracts/Whitelist.sol
@@ -36,10 +36,21 @@ contract Whitelist is Ownable {
         address indexed strike,
         address indexed collateral
     );
+    /// @notice emitted when owner blacklist a product
+    event ProductBlacklisted(
+        bytes32 productHash,
+        address indexed underlying,
+        address indexed strike,
+        address indexed collateral
+    );
     /// @notice emits an event when a collateral address is whitelisted by the owner address
     event CollateralWhitelisted(address indexed collateral);
+    /// @notice emits an event when a collateral address is blacklist by the owner address
+    event CollateralBlacklisted(address indexed collateral);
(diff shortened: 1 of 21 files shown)


contracts/Whitelist.sol
```diff
@@ -105,60 +116,99 @@ contract Whitelist is Ownable {
...
      * @notice set an otoken address as whitelisted
      * @param _otokenAddress address of the oToken that is being whitelisted
+     * @param _state represent if a Otoken is whitelisted or not, true to whitelist and false to blacklist
      */
-    function _setWhitelistedOtoken(address _otokenAddress) internal {
-        require(!whitelistedOtoken[_otokenAddress], "Whitelist: Otoken already whitelisted");
-
-        whitelistedOtoken[_otokenAddress] = true;
+    function _setWhitelistedOtoken(address _otokenAddress, bool _state) internal {
+        whitelistedOtoken[_otokenAddress] = _state;
```
antoncoding, line 212: I don't think we need this function anymore if it's just `whitelistedOtoken[_otokenAddress] = _state;`

antoncoding (approved):
LGTM 🔥

aparnakr (approved):
LGTM!

contracts/Whitelist.sol
haythem96: yeah agree, removed that.

PR #114, opened 2020-09-03
Task/Consider collateral price in calculator

antoncoding:
# Task: Update Calculator

## High Level Description

Add logic to deal with options that have different strike/ underlying than collateral.
Add one more function called `_getToCollateralRate` that deal with this conversion rate

### For put:
* conversion only happen when strike != collateral
* eg, underlying = ETH, strike = USDC, collateral = cUSDC
* in this case, the output of `_getMarginRequired` will be denominated in strike (USDC)
* the output of `_getToCollateralRate` will tell us 1 USDC = ? cUSDC. Let's say ? = 50
* assume output of `_getMarginRequired` is 50 USDC, the real required collateral is 2500 USD.

### For call 
* conversion only happen when underlying != collateral
* eg, underlying = ETH, strike = USDC, collateral = cETH
* in this case, the output of `_getMarginRequired` will be denominated in strike (ETH)
* the output of `_getToCollateralRate` will tell us 1 ETH = ? cETH. Let's say ? = 50
* assume output of `_getMarginRequired` is 1 ETH, the real required collateral is 50 cETH 
### Code

- [x] Unit test 100% coverage

### Documentation

### DIFF
--- a/contracts/MarginCalculator.sol
+++ b/contracts/MarginCalculator.sol
@@ -21,6 +21,7 @@ import {MarginAccount} from "./libs/MarginAccount.sol";
 contract MarginCalculator is Initializable {
     using SafeMath for uint256;
     using FixedPointInt256 for FixedPointInt256.FixedPointInt;
+    using FixedPointInt256 for int256;
     address public addressBook;
 
     function init(address _addressBook) external initializer {
@@ -53,7 +54,7 @@ contract MarginCalculator is Initializable {
     /**
      * @notice returns the net value of a vault in the valid collateral asset for that vault i.e. USDC for puts/ ETH for calls
      * @param _vault the theoretical vault that needs to be checked
-     * @return netValue the amount by which the margin is above or below the required amount.
+     * @return excessCollateral the amount by which the margin is above or below the required amount.
      * @return isExcess true if there's excess margin in the vault. In this case, collateral can be taken out from the vault. False if there is insufficient margin and additional collateral needs to be added to the vault to create the position.
      */
     function getExcessCollateral(MarginAccount.Vault memory _vault) public view returns (uint256, bool) {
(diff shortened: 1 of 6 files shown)


contracts/MarginCalculator.sol
```diff
@@ -72,13 +73,22 @@ contract MarginCalculator is Initializable {
         // Vault contains no short tokens: return collateral value.
         if (_isEmptyAssetArray(_vault.shortOtokens)) return (SignedConverter.intToUint(collateralAmount.value), true);
 
+        // get required margin, denominated in strike or underlying asset
```
aparnakr, line 76: this is a little confusing that for puts it is denominated in strike and call in underlying but the same function returns both.

contracts/MarginCalculator.sol
```diff
@@ -72,13 +73,22 @@ contract MarginCalculator is Initializable {
...
         FixedPointInt256.FixedPointInt memory marginRequirement = _getMarginRequired(_vault);
```
aparnakr, line 77: can we split this up for puts and calls? 

have it as marginRequiredInStrikeAsset = .... 
and marginRequiredInUnderlyingAsset = ... 

if not the naming is a little confusing

aparnakr: Alternatively can all this happen in the getMarginRequired function?

aparnakr (approved):
Looks good, I think the exchange rate calcluation / nomenclature can be simplified. Left some ideas, happy to brainstorm more. It took me a bit to understand, so it might take someone who doesn't know the system inside out a little longer to understand.

the terminology margin and collateral and swapping the two was confusing. Its okay to merge for now and simplify it later (if this is urgent).

haythem96 (approved):

haythem96:
Let's merge this after finishing Controller.

PR #116, opened 2020-09-03
Task/ Calculator scaling

antoncoding:
# Task: Add scaling in Calculator

## High Level Description

* Add 2 more functions to convert token amount into **internal amount**: scaled in 1e18.
  * _tokenAmountToInernalAmount
  * _inernalAmountToTokenAmount

Examples: 
```
USDC decimals = 6
10,000,000 USDC token amount = 10 * 1e18 USDC in internal amount (10 USDC) 
```
* Rename FixedPointInt256 => FPI

### Code

- [x] Unit test 100% coverage

### Documentation


contracts/MarginCalculator.sol
```diff
@@ -63,30 +64,40 @@ contract MarginCalculator is Initializable {
         // ensure the long asset is valid for the short asset.
         require(_isMarginableLong(_vault), "MarginCalculator: long asset not marginable for short asset");
```
aparnakr, line 66: need a check to make sure collateral in vault == short.collateral if there is a collateral asset

contracts/MarginCalculator.sol
```diff
@@ -350,4 +350,51 @@ contract MarginCalculator is Initializable {
...
+     */
+
+    function _tokenAmountToInernalAmount(uint256 _amount, address _token) internal view returns (uint256) {
+        ERC20Interface token = ERC20Interface(_token);
+        uint256 decimals = uint256(token.decimals());
+        uint256 base = 18;
+        if (decimals == base) return _amount;
+        if (decimals > base) {
+            uint256 exp = decimals - base;
+            return _amount.div(10**exp);
```
aparnakr, line 371: isn't this a safe math div? So doesn't this do 

8000000 / 10^6? Where is the scaling by 1e18 happening?

aparnakr: Also you should probs first multiply by 1e18, since if you pass in something like 80 for USDC, it will show 0 in internal values right?

test/marginCalculator.test.ts
```diff
@@ -520,66 +545,112 @@ contract('MarginCalculator', () => {
...
+
+      it('(3) Short: 1 300 put, collateral: 350 rUSDC => excess: 50', async () => {
+        const collateralAmount = createTokenAmount(350, rtokenDecimals)
+        const expectOutPut = createTokenAmount(50, rtokenDecimals)
+        const vault = createVault(put.address, undefined, rusd.address, amountOne, undefined, collateralAmount)
+        const [netValue, isExcess] = await calculator.getExcessCollateral(vault)
+        assert.equal(isExcess, true)
+        assert.equal(netValue.toString(), expectOutPut.toString())
       })
     })
```
aparnakr, line 594: To make sure the decimals actually work you should test something like 1.5 or 0.0004 USDC. If you test big USDC values, it won't run into the rounding to 0 issue.

aparnakr (changes requested):

contracts/MarginCalculator.sol
antoncoding: this case is if the USDC decimals is higher than 18. Let's say there's a token has 20 decimals and you pass in 20, then yes it will be truncated and the output will be 0.

antoncoding: if the token has decimals 20, it already scaled by 1e20 in our definition, we just have to remove the last 2 digits to get a number scaled in 1e18.

test/marginCalculator.test.ts
antoncoding: it's not an issue, it's an expected behavior to truncated decimals more than 20. But yeah just add a test for that

contracts/MarginCalculator.sol
antoncoding: GOOD CATCH!

aparnakr (approved):
LGTM~!

contracts/MarginCalculator.sol
```diff
@@ -355,8 +355,8 @@ contract MarginCalculator is Initializable {
      * @dev convert uint256 to FixedPointInt, no scaling invloved
      * @return the FixedPointInt format of input
      */
-    function _uint256ToFixedPointInt(uint256 _num) internal pure returns (FixedPointInt256.FixedPointInt memory) {
-        return FixedPointInt256.FixedPointInt(SignedConverter.uintToInt(_num));
+    function _uint256ToFPI(uint256 _num) internal pure returns (FPI.FixedPointInt memory) {
```
haythem96, line 358: This function is not doing any scaling, right ?

antoncoding: no, I only rename it to a shorter name

contracts/MarginCalculator.sol
```diff
@@ -366,4 +366,55 @@ contract MarginCalculator is Initializable {
...
+     * Examples:
+     * (1)  USDC    decimals = 6
+     *      Input:  8 * 1e18 (8.0 USDC)   =>   Output:  8000000 USDC
+     * (2)  cUSDC   decimals = 8
+     *      Input:  8 * 1e16 (0.08 cUSDC) =>   Output:  8000000 cUSDC
+     * (3)  rUSD    decimals = 20 (random USD)
+     *      Input:  1                    =>    Output:  100     rUSDC
+     * @return token amount in its native form.
+     */
+    function _inernalAmountToTokenAmount(uint256 _amount, address _token) internal view returns (uint256) {
```
haythem96, line 407: fix typo `_internalAmountToTokenAmount`

haythem96 (approved):

PR #120, opened 2020-09-04
Integration test: Chainlink Pricer

antoncoding:
# Task: Feature Name:

## High Level Description

Specific Changes
Function x was added ...

### Code

- [ ] Unit test 100% coverage

### Documentation

### DIFF
--- a/test/integration-tests/chainlinkPricer-oracle.test.ts
+++ b/test/integration-tests/chainlinkPricer-oracle.test.ts
@@ -1,0 +1,171 @@
+import {
+  ChainLinkPricerInstance,
+  OracleInstance,
+  MockChainlinkAggregatorInstance,
+  MockERC20Instance,
+} from '../../build/types/truffle-types'
+import BigNumber from 'bignumber.js'
+
+const {expectRevert, time} = require('@openzeppelin/test-helpers')
+
+const ChainlinkPricer = artifacts.require('ChainLinkPricer.sol')
+const Oracle = artifacts.require('Oracle.sol')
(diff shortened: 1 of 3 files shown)


test/integration-tests/chainlinkPricer-oracle.test.ts
```diff
@@ -0,0 +1,171 @@
...
+  MockChainlinkAggregatorInstance,
+  MockERC20Instance,
+} from '../../build/types/truffle-types'
+import BigNumber from 'bignumber.js'
+
+const {expectRevert, time} = require('@openzeppelin/test-helpers')
+
+const ChainlinkPricer = artifacts.require('ChainLinkPricer.sol')
+const Oracle = artifacts.require('Oracle.sol')
+const MockChainlinkAggregator = artifacts.require('MockChainlinkAggregator.sol')
```
aparnakr, line 13: Can you use the actual chainLink contract instead of the aggregator?

aparnakr (approved):
LGTM

test/integration-tests/chainlinkPricer-oracle.test.ts
antoncoding: I think using Mock make more sense in this case!
In tests we mock things that's out of our control. In this case, aggregator is not part of our system and we won't be able to directly write to the aggregator when we're live on mainnet, so I don't think that's should be included in the test.
Everything we care about is reading from the aggregator, and this is the perfect situation of using mock contracts. If we copy paste Chainlink aggregator, we will need like 10 steps  to submit a price to the aggregator, and we can't do the same thing on Mainnet either

PR #119, opened 2020-09-04
Burn otoken functionality

haythem96:
# Task: Feature Name:

Controller: burn otoken 

## High Level Description

This PR add the functionality of burning un-expired short otoken.

### Code

- [x] Unit test 100% coverage
- [x] Does your code follow the naming and code documentation guidelines?

### Documentation

- [x] Is your code up to date with the spec? 
- [x] Have you added your tests to the testing doc?

PR #128, opened 2020-09-05
Exercise functionality

haythem96:
# Task: Feature Name:

Controller module: exercise functionality

## High Level Description

This PR add the exercise functionality in Controller module.
Should be reviewed/merged after #119 .

### Code

- [x] Unit test 100% coverage

### Documentation


contracts/Controller.sol
```diff
@@ -495,13 +508,64 @@ contract Controller is ReentrancyGuard, Ownable {
...
+     * @param _amount amount of Otoken
+     * @return payout = cashValue * amount
+     */
+    function getPayout(address _otoken, uint256 _amount) internal view returns (uint256) {
+        address calculatorModule = AddressBookInterface(addressBook).getMarginCalculator();
+        MarginCalculatorInterface calculator = MarginCalculatorInterface(calculatorModule);
+
+        uint256 cashValue = calculator.getExpiredCashValue(_otoken);
+
+        return cashValue.mul(_amount).div(1e18);
```
aparnakr, line 567: Is the expiredCashValue always going to be 1e18 or is that dependent on the token?

aparnakr: @antoncoding

test/controller.test.ts
```diff
@@ -1981,6 +1966,536 @@ contract('Controller', ([owner, accountOwner1, accountOperator1, random]) => {
...
+          sender: holder1,
+          asset: shortOtoken.address,
+          vaultId: '0',
+          amount: shortAmountToBurn.toNumber(),
+          index: '0',
+          data: ZERO_ADDR,
+        },
+      ]
+
+      assert.equal(await controller.isExpired(shortOtoken.address), true, 'Short otoken is not expired yet')
```
aparnakr, line 2442: This message should be price not finalized

test/controller.test.ts
```diff
@@ -2052,6 +2567,17 @@ contract('Controller', ([owner, accountOwner1, accountOperator1, random]) => {
...
+        weth.address,
+        usdc.address,
+        usdc.address,
+        new BigNumber(200).times(new BigNumber(10).exponentiatedBy(18)),
+        new BigNumber(await time.latest()).plus(60000 * 60000),
+        true,
+      )
+
       assert.equal(await controller.isExpired(otoken.address), false, 'Otoken expiry check mismatch')
     })
```
aparnakr, line 2583: Can you add a test to be able to exercise many series of oTokens in 1 transaction?

aparnakr (commented):
Once we figure out the math and add the tests we can merge.

antoncoding:
image

contracts/Controller.sol
antoncoding: always 1e18

contracts/Controller.sol
```diff
@@ -495,13 +508,64 @@ contract Controller is ReentrancyGuard, Ownable {
...
+        emit Exercise(_args.otoken, msg.sender, _args.exerciser, otoken.collateralAsset(), _args.amount, payout);
+    }
+
+    /**
+     * @notice get Otoken payout after expiry
+     * @param _otoken Otoken address
+     * @param _amount amount of Otoken
+     * @return payout = cashValue * amount
+     */
+    function getPayout(address _otoken, uint256 _amount) internal view returns (uint256) {
```
antoncoding, line 561:
```suggestion
    function _getPayout(address _otoken, uint256 _amount) internal view returns (uint256) {
```

antoncoding: or consider making this public?

contracts/Controller.sol
```diff
@@ -93,6 +93,23 @@ contract Controller is ReentrancyGuard, Ownable {
         uint256 vaultId,
         uint256 amount
     );
+    /// @notice emits an event when a short otoken get burned from a vaukt
```
antoncoding, line 96:
```suggestion
    /// @notice emits an event when a short otoken get burned from a vault
```

contracts/Controller.sol
```diff
@@ -93,6 +93,23 @@ contract Controller is ReentrancyGuard, Ownable {
...
+        address indexed AccountOwner,
+        address indexed from,
+        uint256 vaultId,
+        uint256 amount
+    );
+    /// @notice emits an event when a exercise action execute
+    event Exercise(
+        address indexed otoken,
+        address indexed from,
+        address indexed exerciser,
```
antoncoding, line 108: Currently these 2 means msg.sender and exerciser and it could be quite confusing. Maybe change the name exerciser to owner?

contracts/Controller.sol
```diff
@@ -495,13 +508,64 @@ contract Controller is ReentrancyGuard, Ownable {
...
      * @param _args MintArgs structure
      */
-    // function _burnOtoken(Actions.BurnArgs memory _args) internal {}
+    function _burnOtoken(Actions.BurnArgs memory _args)
+        internal
+        isAuthorized(msg.sender, _args.owner)
+        returns (MarginAccount.Vault memory)
+    {
+        require(checkVaultId(_args.owner, _args.vaultId), "Controller: invalid vault id");
+        require(_args.from == msg.sender, "Controller: burner address and msg.sender address mismatch");
```
antoncoding, line 517: I think we can freely burn oToken from the owner's account without having to move tokens to msg.sender first, because the payout always go to the owner so there's nothing to worry about.

contracts/libs/Actions.sol
```diff
@@ -246,6 +246,7 @@ library Actions {
      */
     function _parseExerciseArgs(ActionArgs memory _args) internal pure returns (ExerciseArgs memory) {
         require(_args.actionType == ActionType.Exercise, "Actions: can only parse arguments for exercise actions");
+        require(_args.sender != address(0), "Actions: cannot exercise to an invalid account");
 
         return ExerciseArgs({exerciser: _args.sender, otoken: _args.asset, amount: _args.amount});
     }
```
antoncoding, line 252: Please add test for this

contracts/mocks/MockMarginCalculator.sol
```diff
@@ -206,7 +211,10 @@ contract MockMarginCalculator {
     }
 
     function _getUnderlyingPrice(address _otoken) internal view returns (uint256, bool) {
-        return (price[_otoken], isPriceFinalized[_otoken]);
+        OracleInterface oracle = OracleInterface(AddressBookInterface(addressBook).getOracle());
+        OtokenInterface otoken = OtokenInterface(_otoken);
+
+        return oracle.getExpiryPrice(otoken.underlyingAsset(), otoken.expiryTimestamp());
```
antoncoding, line 217: Why using the oracle contract? 
This will break other tests.

antoncoding: I think we should keep the mock contract simple, why adding oracle into the dependency graph

test/controller.test.ts
```diff
@@ -1981,6 +1966,536 @@ contract('Controller', ([owner, accountOwner1, accountOperator1, random]) => {
...
+          )
+        })
+      })
+    })
+  })
+
+  describe('Exercise', () => {
+    let shortOtoken: MockOtokenInstance
+
+    before(async () => {
```
antoncoding, line 2290: separate into before and an `it` block. If you have assert.equal then you're testing something, you should put that in a `it` block

antoncoding (changes requested):
Add test, update MockCalculator usage

contracts/Controller.sol
haythem96: why ? can be calculated by calling the getCashValue which is public and multiplying by the amount.

contracts/mocks/MockMarginCalculator.sol
haythem96: It is is simple, It just call another Mock contract to get price.
I made that change because the exercise function use the Oracle mock, the the `_getUnderlyingPrice` was using a mapping to mock prices, so I needed to push price+finalized in two mocks for exercise to work.
Better to have both of them using same Oracle mock. Also this does not break any tests.

test/controller.test.ts
haythem96: mmm this assert check that the Otoken is expired! can u explain more ?

contracts/Controller.sol
haythem96: Updated the names in the event+parse args.

contracts/libs/Actions.sol
haythem96: added.

contracts/Controller.sol
antoncoding: just thought i could be handy but don't need to. 
if you want to keep it internal then add a `_`

test/controller.test.ts
haythem96: added.

haythem96:
> Add test, update MockCalculator usage

Added tests, will keep mock as it is.

antoncoding (approved):

aparnakr (approved):

PR #139, opened 2020-09-07
Refactor id and vault validation

antoncoding:
# Task: Refactor id and vault validation

### Feature added
* Add vault owner check on all _runActions

### Reduce contract size:
* refactor modifier according to [this post](https://blog.polymath.network/solidity-tips-and-tricks-to-save-gas-and-reduce-bytecode-size-c44580b218e6)
* Simplify

### Gas optimization

Gas usage before:

**Gas usage before**
min gas 51988
max gas 358170
avg147701

 
**Refactor _checkActionsVaults**
min gas 51946
max gas 358040
avg 147649

**remove return vault from _mintOtoken, _burnOtoken...etc**
min gas 52526
max gas 345815
avg 143227

**Final gas (include checking vault owners, update modifiers)**
min gas 52647
max gas 346216
avg 143516

### Code

- [x] Unit test 100% coverage

### Documentation


contracts/libs/MarginAccount.sol
```diff
@@ -199,17 +199,4 @@ library MarginAccount {
...
-     * @param _vault The vault that the user is clearing.
-     */
-    function _clearVault(Vault storage _vault) internal {
-        delete _vault.shortAmounts;
-        delete _vault.longAmounts;
-        delete _vault.collateralAmounts;
-        delete _vault.shortOtokens;
-        delete _vault.longOtokens;
-        delete _vault.collateralAssets;
-    }
```
aparnakr, line 214: can you separate these from the controller changes?

aparnakr (approved):
The refactor of controller is good, the delete vault part needs some fixes. Can you separate the two and merge the controller refactor?

contracts/Controller.sol
```diff
@@ -273,12 +285,21 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
...
+        internal
+        returns (
+            bool,
+            address,
+            uint256
+        )
+    {
+        address vaultOwner;
+        uint256 vaultId; // 0
+        bool vaultUpdated; // false
```
haythem96, line 303: remove the comment `//`

contracts/Controller.sol
```diff
@@ -116,24 +116,36 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
...
+     */
+    function _isAuthorized(address _accountOwner) internal view {
         require(
-            (_sender == _accountOwner) || (operators[_accountOwner][_sender]),
+            (msg.sender == _accountOwner) || (operators[_accountOwner][msg.sender]),
             "Controller: msg.sender is not authorized to run action"
         );
-
-        _;
     }
```
haythem96, line 149: Why this modification ?

antoncoding: save some bytes for passing in that parameter

haythem96: Would like to keep it with two arguments, it much clearer that way, unless it is saving a lot of bytes...

haythem96: But also why having a modifier calling the function ?

antoncoding: because if it's a modifier, it copy all the code in the modifier into the function body.

antoncoding: So if we have multiple function with the same modifier, better use an internal function in the modifier, so the same code won't be copied multiple times

antoncoding: I'll make it back to using 2 arguments then

haythem96 (approved):

PR #141, opened 2020-09-07
Refactor: Use internal state variables

antoncoding:
# Task: Feature Name:

## High Level Description

Makes 6 state variables internal
This saves 50 bytes

### Code

- [x] Unit test 100% coverage

### Documentation


contracts/Controller.sol
```diff
@@ -27,11 +27,11 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
...
-    AddressBookInterface public addressbook;
-    WhitelistInterface public whitelist;
-    OracleInterface public oracle;
-    MarginCalculatorInterface public calculator;
-    MarginPoolInterface public pool;
+    AddressBookInterface internal addressbook;
+    WhitelistInterface internal whitelist;
+    OracleInterface internal oracle;
+    MarginCalculatorInterface internal calculator;
+    MarginPoolInterface internal pool;
```
aparnakr, line 34: hmm imo still valuable to have these as public. During upgrades it could get confusing if we can't see the state changes properly.

aparnakr (commented):
Not sure about these changes. This data is still valuable for us to be able to see when we make upgrades

contracts/Controller.sol
haythem96: @antoncoding Can u add a getter for those ? like `getConfiguration()`

haythem96 (approved):

aparnakr (approved):

PR #145, opened 2020-09-08
Fix Payout calculation

antoncoding:
# Task: Update Exercise payout calculation
## High Level Description

* Fix payout calculation
* Refactor Controller and Calculator to inherit the same OMath contract.
* Refactor controller test cases to avoid using small numbers

### Code

- [x] Unit test 100% coverage

### Documentation


contracts/Controller.sol
```diff
@@ -5,6 +5,7 @@ pragma solidity =0.6.10;
 
 pragma experimental ABIEncoderV2;
 
+import {OMath} from "./OMath.sol";
```
haythem96, line 8: Can we change the `OMath` naming ? like `Omath` or something... just one uppercase letter in the beginning.

haythem96 (approved):

contracts/OMath.sol
```diff
@@ -0,0 +1,63 @@
...
+ * SPDX-License-Identifier: UNLICENSED
+ */
+pragma solidity 0.6.10;
+pragma experimental ABIEncoderV2;
+
+import {SafeMath} from "./packages/oz/SafeMath.sol";
+import {OtokenInterface} from "./interfaces/OtokenInterface.sol";
+import {ERC20Interface} from "./interfaces/ERC20Interface.sol";
+
+contract OMath {
```
aparnakr, line 11: hmm can we call this something else? 

Some ideas: 
Converter
TokenTo1e18Converter

contracts/Controller.sol
```diff
@@ -614,13 +615,29 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
...
+        if (strike == collateral) {
+            payoutAmount = amountInStrike;
+        } else {
+            uint256 expiry = otoken.expiryTimestamp();
+            (uint256 strikePrice, ) = oracle.getExpiryPrice(strike, expiry);
+            (uint256 collateralPrice, ) = oracle.getExpiryPrice(collateral, expiry);
+            payoutAmount = amountInStrike.mul(strikePrice).div(collateralPrice);
+        }
+
+        return _internalAmountToTokenAmount(payoutAmount, collateral);
```
aparnakr, line 640: wait alternatively, can we have this calculation in calculator? Then we don't need oMath at all right?

contracts/Controller.sol
```diff
@@ -614,13 +615,29 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
...
 
-        return cashValue.mul(_amount).div(1e18);
+        OtokenInterface otoken = OtokenInterface(_otoken);
+        address strike = otoken.strikeAsset();
+        address collateral = otoken.collateralAsset();
+
+        uint256 payoutAmount;
+
+        if (strike == collateral) {
+            payoutAmount = amountInStrike;
```
aparnakr, line 632: even this will fail the integration test right, because for calls the getExpiredCashValue is in USD, but the collateral is in ETH. So where does the USD to ETH price conversion happen?

aparnakr (changes requested):
Is this trying to fix the issue the integration test was running into? if so, I think that issue is not fixed.

aparnakr (approved):
LGTM

contracts/Controller.sol
antoncoding: we can't have it in the getExpiredCashValue function, but we can use a separate function for that.

contracts/MarginCalculator.sol
```diff
@@ -35,24 +35,76 @@ contract MarginCalculator {
...
+        // divide price of underlying by price of strike,
+        // to get the real price of underlying denominated in strike at expiry
+        (uint256 underlyingPrice, bool isUnderlyingFinalized) = _getAssetPrice(
+            otoken.underlyingAsset(),
+            otoken.expiryTimestamp()
+        );
+        (uint256 strikeAssetPrice, bool isStrikeFinalized) = _getAssetPrice(
+            otoken.strikeAsset(),
+            otoken.expiryTimestamp()
+        );
```
aparnakr, line 57: so for usdc are we returning 1e18? What happens if the asset doesn't have an oracle?

contracts/MarginCalculator.sol
```diff
@@ -35,24 +35,76 @@ contract MarginCalculator {
      * @dev For call return = Max (0, ETH Price - oToken.strike)
      * @dev For put return Max(0, oToken.strike - ETH Price)
      * @param _otoken otoken address
-     * @return the cash value of an expired otoken
+     * @return the cash value of an expired otoken, denomincated in strike asset. scaled by 1e18
```
aparnakr, line 38:
```suggestion
     * @return the cash value of an expired otoken, denominated in strike asset. scaled by 1e18
```

contracts/MarginCalculator.sol
aparnakr: Also if strike == underlying, does this run into the rounding error you were talking about earlier?

contracts/MarginCalculator.sol
```diff
@@ -35,24 +35,76 @@ contract MarginCalculator {
...
+        uint256 exchangeRate;
+
+        if (strike == collateral) {
+            exchangeRate = cashValueInStrike;
+        } else {
+            uint256 expiry = otoken.expiryTimestamp();
+            // isFinalized of strike asset is already checked in _getExpiredCashValue
+            (uint256 strikePrice, ) = _getAssetPrice(strike, expiry);
+            (uint256 collateralPrice, bool isCollateralPriceFinalized) = _getAssetPrice(collateral, expiry);
+            require(isCollateralPriceFinalized, "MarginCalculator: collateral price is not finalized");
```
aparnakr, line 99: don't you also need strikePrice to be finalized?

contracts/MarginCalculator.sol
```diff
@@ -138,15 +190,15 @@ contract MarginCalculator {
                 marginRequired = _getCallSpreadMarginRequired(shortAmount, longAmount, shortStrike, longStrike);
             }
         } else {
-            FPI.FixedPointInt memory shortCashValue = _uint256ToFPI(getExpiredCashValue(address(short)));
+            FPI.FixedPointInt memory shortCashValue = _uint256ToFPI(_getExpiredCashValue(address(short)));
             FPI.FixedPointInt memory longCashValue = hasLongInVault
-                ? _uint256ToFPI(getExpiredCashValue(_vault.longOtokens[0]))
+                ? _uint256ToFPI(_getExpiredCashValue(_vault.longOtokens[0]))
```
aparnakr, line 195: shouldn't the getExcessCollateral Function return the value in the collateral asset? So shouldn't this call getExpiredPayoutRate?

test/controller.test.ts
```diff
@@ -1601,16 +1606,16 @@ contract('Controller', ([owner, accountOwner1, accountOwner2, accountOperator1,
         const vaultCounter = new BigNumber(await controllerProxy.getAccountVaultCounter(accountOwner1))
         assert.isAbove(vaultCounter.toNumber(), 0, 'Account owner have no vault')
 
-        const collateralToDeposit = new BigNumber(await shortOtoken.strikePrice()).dividedBy(1e18)
-        const amountToMint = new BigNumber('1')
+        const collateralToDeposit = new BigNumber(await shortOtoken.strikePrice()).dividedBy(1e12)
+        const amountToMint = new BigNumber(1e18)
```
aparnakr, line 1610: instead of hardcoding these 1e18, 1e12, can you have it calculate it based on the token's decimals? Eg if collateral is USDC, then use usdc.decimals?

aparnakr (changes requested):
The getExcessCollateral function still seems to behave differently before and after expiry. Before expiry its returning the amount collateral needed for call options in ETH, after expiry it seems to be giving the amount of excess collateral denominated in USD / strike asset.

contracts/MarginCalculator.sol
antoncoding: no, all the formula we had were based on cash value in strike, and the getExpiredCashValue function remains the same

antoncoding:
> The getExcessCollateral function still seems to behave differently before and after expiry. Before expiry its returning the amount collateral needed for call options in ETH, after expiry it seems to be giving the amount of excess collateral denominated in USD / strike asset.

I don't think that's the case:
In `_getMarginRequired`, it's basically our formula for put / calls before and after expiry. For puts, it returns value in strike, (USDC), and for calls, it returns value in underlying (ETH), not collateral.

So In 91~94
```javascript
        // get required margin, denominated in strike or underlying asset
        FPI.FixedPointInt memory marginRequirement = _getMarginRequired(_vault);
        // get exchange rate to convert marginRequirement to amount of collateral
        FPI.FixedPointInt memory exchangeRate = _getToCollateralRate(_vault.shortOtokens[0]);
```
after I get the `marginRequirement` which can be denominated in strike or underlying, I used another function to get the assets's exchange rate to convert it into amount of collateral. For puts, this function gets underlying => collateral rate, for calls, it gets strike to collateral rate.

contracts/MarginCalculator.sol
antoncoding: isFinalized of strike asset is already checked in _getExpiredCashValue

contracts/MarginCalculator.sol
antoncoding: yeah, for usdc we will need strikeAssetPrice to be 1e18.
We will need 3 oracles for expiry: strike / underlying and collateral. For the most cases, strike or underlying will = collateral, unless we want to do farming option (cETH as collateral...etc)

antoncoding: Also re rounding problem, it seems to only happen with `mul`, because all the old test cases passed. I'll need to investigate more into that.

aparnakr (approved):

aparnakr:
> > The getExcessCollateral function still seems to behave differently before and after expiry. Before expiry its returning the amount collateral needed for call options in ETH, after expiry it seems to be giving the amount of excess collateral denominated in USD / strike asset.
> 
> I don't think that's the case:
> In `_getMarginRequired`, it's basically our formula for put / calls before and after expiry. For puts, it returns value in strike, (USDC), and for calls, it returns value in underlying (ETH), not collateral.
> 
> So In 91~94
> 
> ```js
>         // get required margin, denominated in strike or underlying asset
>         FPI.FixedPointInt memory marginRequirement = _getMarginRequired(_vault);
>         // get exchange rate to convert marginRequirement to amount of collateral
>         FPI.FixedPointInt memory exchangeRate = _getToCollateralRate(_vault.shortOtokens[0]);
> ```
> 
> after I get the `marginRequirement` which can be denominated in strike or underlying, I used another function to get the assets's exchange rate to convert it into amount of collateral. For puts, this function gets underlying => collateral rate, for calls, it gets strike to collateral rate.

That makes sense. I got confused there. My bad. We should definitely try to simplify the margin calculator more 😅

test/controller.test.ts
```diff
@@ -3456,23 +3547,12 @@ contract('Controller', ([owner, accountOwner1, accountOwner2, accountOperator1,
...
-        true,
-      )
-      await oracle.setIsDisputePeriodOver(
-        await shortOtoken.underlyingAsset(),
-        new BigNumber(await shortOtoken.expiryTimestamp()),
-        true,
-      )
+      const expiry = new BigNumber(await shortOtoken.expiryTimestamp())
+      await oracle.setExpiryPriceFinalizedAllPeiodOver(weth.address, expiry, new BigNumber(150).times(1e18), true)
+      await oracle.setExpiryPriceFinalizedAllPeiodOver(usdc.address, expiry, new BigNumber(1).times(1e18), true)
```
haythem96, line 3552: I think it is better to not use this function as it doesn't not exist in Oracle module, I understand that this mock function is mimicking the same behavior, but better to follow same flow IMO.

contracts/MarginCalculator.sol
```diff
@@ -104,6 +111,51 @@ contract MarginCalculator {
...
 
+    /**
+     * @notice Return the cash value of an expired oToken.
+     * @dev For call return = Max (0, ETH Price - oToken.strike)
+     * @dev For put return Max(0, oToken.strike - ETH Price)
+     * @param _otoken otoken address
+     * @return the cash value of an expired otoken, denominated in strike asset. scaled by 1e18
+     */
+    function _getExpiredCashValue(address _otoken) internal view returns (uint256) {
+        require(_otoken != address(0), "MarginCalculator: Invalid token address.");
```
haythem96, line 122: This require statement can be moved in `getExpiredPayoutRate()`

contracts/MarginCalculator.sol
```diff
@@ -32,26 +32,33 @@ contract MarginCalculator {
...
      * @param _otoken otoken address
-     * @return the cash value of an expired otoken
+     * @return the exchange rate that shows how much collateral unit can be take out by 1 otoken unit, scaled by 1e18
      */
-    function getExpiredCashValue(address _otoken) public view returns (uint256) {
-        require(_otoken != address(0), "MarginCalculator: Invalid token address.");
+    function getExpiredPayoutRate(address _otoken) external view returns (uint256) {
+        uint256 cashValueInStrike = _getExpiredCashValue(_otoken);
+
         OtokenInterface otoken = OtokenInterface(_otoken);
```
haythem96, line 42: We can move this casting in the first line of this function, and pass `OtokenInterface` as argument also to `_getExpiredCashValue()` as it cast the address there also.

PR #136, opened 2020-09-06
Emergency shutdown implementation

haythem96:
# Emergency shutdown implementation

## High Level Description

This PR implement the emergency shutdown mechanism.
Closes #91 .

### Code

- [x] Unit test 100% coverage
- [x] Does your code follow the naming and code documentation guidelines?

### Documentation

- [x] Is your code up to date with the spec? 
- [x] Have you added your tests to the testing doc?

PR #148, opened 2020-09-08
Call action implementation

haythem96:
# Call action implementation

## High Level Description

This PR implement the call action functionality. 
- Call action can execute any arbitrary call to any address.
- Call action forward the msg.value of the transaction equal to the amount filed.

Should be reviewed/merged after #136 .

### Code

- [x] Unit test 100% coverage

### Documentation


haythem96:
The `callFunction()` have the vault owner and vault id as arguments, to allow for implementation that need access to those data for any kind of verification/process.

contracts/external/wrapper/FlashWrap.sol
```diff
@@ -0,0 +1,38 @@
...
+pragma experimental ABIEncoderV2;
+
+import {WETH9} from "../../packages/canonical-weth/WETH9.sol";
+
+/**
+ * @author Opyn Team
+ * @title FlashWrap
+ * @notice contract To wrap ETH
+ */
+contract FlashWrap {
```
antoncoding, line 15:
```suggestion
contract FlashWrap is CalleeInterface {
```
To make it more clear

antoncoding (approved):
Need to exclude the emergency shutdown implementation to merge.

antoncoding:
Is there a way that we can make sure user don't accidentally send eth to us in `operate`?

aparnakr (approved):
Looks good to me! Can we in the future also add a test to make sure we can do a trade on uniswap / balancer? (not urgent now)

haythem96:
> Looks good to me! Can we in the future also add a test to make sure we can do a trade on uniswap / balancer? (not urgent now)

We can do it in the integration tests.

aparnakr:
BTW are you adding the whitelist call actions address functionality here or in a different PR?

haythem96:
> BTW are you adding the whitelist call actions address functionality here or in a different PR?

Will add it here.

contracts/libs/Actions.sol
```diff
@@ -275,6 +281,13 @@ library Actions {
         require(_args.actionType == ActionType.Call, "Actions: can only parse arguments for call actions");
         require(_args.sender != address(0), "Actions: target address cannot be address(0)");
 
-        return CallArgs({owner: _args.owner, callee: _args.sender, data: _args.data});
+        return
+            CallArgs({
+                owner: _args.owner,
+                callee: _args.sender,
+                vaultId: _args.vaultId,
+                msgValue: _args.amount,
```
antoncoding, line 289: I think we should call it value instead of `msgValue`. message value implies that this is the same as msg.value, but it doesn't. If a user pass in 2 call actions each with value = 5, the msg.value will need to be 10. This may get quite confusing.

PR #151, opened 2020-09-08
Chainlink Pricer: Fix price not scaled for expiryPrice

antoncoding:
# Task: Feature Name:

## High Level Description

Fix price not scaled for expiryPrice

### Code

- [x] Unit test 100% coverage

### Documentation

### DIFF
--- a/contracts/pricers/ChainlinkPricer.sol
+++ b/contracts/pricers/ChainlinkPricer.sol
@@ -12,6 +12,8 @@ import {SafeMath} from "../packages/oz/SafeMath.sol";
 contract ChainLinkPricer is OpynPricerInterface {
     using SafeMath for uint256;
 
+    uint256 internal constant BASE = 1e10;
+
     /// @notice the opyn oracle address
     OracleInterface public oracle;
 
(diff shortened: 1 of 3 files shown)


contracts/pricers/ChainlinkPricer.sol
```diff
@@ -45,8 +47,8 @@ contract ChainLinkPricer is OpynPricerInterface {
     function getPrice() external override view returns (uint256) {
         int256 answer = aggregator.latestAnswer();
         require(answer > 0, "ChainLinkPricer: price is lower than 0");
-        // Scaled to 1e18 from 1e8
-        return uint256(answer * 1e10);
+        // Scale to 1e18 from 1e8
+        return uint256(answer).mul(BASE);
```
haythem96, line 51: Can we start using the FixedPoint to represent any scaled variable ? instead of uint256

aparnakr: +1

aparnakr: Is the chainLink on chain prices 1e8?

antoncoding: don't think we need to because our FixPoint is int256.
Also that will change everything in Oracle, can do that in a separate PR as improvement

haythem96 (approved):

aparnakr (approved):

PR #152, opened 2020-09-09
MarginPool: Add require check for ERC20 operations

### LINKED ISSUE #149: update MarginPool and tests for standard ERC20 transfers
In the current implementation, we ignore the returned value of transfer and transferFrom. In the ERC20 standard, a failed transfer should return false instead of revert.

May need to do some more research on if there're any tokens that won't return true, and look into other protocols on how they handle it.

antoncoding:
# Task: Add require check in MarginPool

closes #149 

### Code

- [x] Unit test 100% coverage

### Documentation


haythem96 (approved):
LGTM. Just change `getStoredBalance()` to `external`.

aparnakr (approved):

PR #153, opened 2020-09-09
Update product hash to include isPut

### LINKED ISSUE #146: Update productID to include isPut in the hash
This will prevent launching an ETH-USDC-USDC call option with our first version.

antoncoding:
# Task: Feature Name:

## High Level Description

Fix #146 

### Todos

- [x] Unit test 100% coverage

### DIFF
--- a/contracts/OtokenFactory.sol
+++ b/contracts/OtokenFactory.sol
@@ -68,7 +68,12 @@ contract OtokenFactory is OtokenSpawner {
 
         address whitelist = AddressBookInterface(addressBook).getWhitelist();
         require(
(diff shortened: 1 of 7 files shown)


aparnakr (approved):
LGTM. Are there any other implications of this PR? why didn't we have this in the first place?

antoncoding:
> LGTM. Are there any other implications of this PR? why didn't we have this in the first place?

In the first iteration, we disallowed puts to have collateral != underlying, and disallowed calls to have collateral != strike. But since we need the yield farming implementation, that constraint was removed.

This will results in allowing users to create USDC-ETH calls with USDC collateral, with the collateral ratio of 1.

haythem96 (approved):

PR #156, opened 2020-09-09
Improve margin pool

haythem96:
# Improve margin pool

## High Level Description

This PR:
- Add SafeERC20 library
- Remove require statement on 0 amount transfer from pool to user.

### Code

- [x] Unit test 100% coverage

### Documentation

### DIFF
--- a/test/controller.test.ts
+++ b/test/controller.test.ts
@@ -671,29 +671,6 @@ contract('Controller', ([owner, accountOwner1, accountOwner2, accountOperator1, 
         )
       })
 
-      it('should revert withdrawing long otoken amount equal to zero', async () => {
-        const vaultCounter = new BigNumber(await controllerProxy.getAccountVaultCounter(accountOwner1))
-        assert.isAbove(vaultCounter.toNumber(), 0, 'Account owner have no vault')
-
-        const actionArgs = [
-          {
-            actionType: ActionType.WithdrawLongOption,
-            owner: accountOwner1,
-            sender: accountOwner1,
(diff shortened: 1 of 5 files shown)


test/controller.test.ts
```diff
@@ -671,29 +671,6 @@ contract('Controller', ([owner, accountOwner1, accountOwner2, accountOperator1,
         )
       })
 
-      it('should revert withdrawing long otoken amount equal to zero', async () => {
-        const vaultCounter = new BigNumber(await controllerProxy.getAccountVaultCounter(accountOwner1))
-        assert.isAbove(vaultCounter.toNumber(), 0, 'Account owner have no vault')
```
aparnakr, line 676: wait can we still have this check in the controller?

aparnakr (approved):
LGTM, added a couple of comments

test/controller.test.ts
haythem96: This check is totally unneeded, will just consume more gas for the user that is withdrawing more than zero amount.

antoncoding (dismissed):
Haven't see you applied the SafeERC20 library to the margin pool, is this intentionally? 

From [Library](https://docs.openzeppelin.com/contracts/3.x/api/token/erc20)
To use this library you can add a using SafeERC20 for IERC20; statement to your contract, which allows you to call the safe operations as token.safeTransfer(…​), etc.

I think we can simply include that in the same PR!

antoncoding (approved):
I saw the thread on Slack. So this is just for removing the constraint, looks good to me

PR #285, opened 2020-11-17
[L02] Constants are not being declared explicitly

antoncoding:
# Task: Fix L02:

There are several occurrences of literal values with unexplained meaning. This PR added the followings:
1. Define a constant `MAX_EXPIRY ` instead of hardcoding `require(_expiry < 11865398400)`
2. Add comment to explain the UTC 08:00 constraint 

Feel free to push commits to this PR if there're other hardcoded number that would better be defined as constant.

### Todos

- [x] Unit test 100% coverage

### DIFF
--- a/contracts/MarginCalculator.sol
+++ b/contracts/MarginCalculator.sol
@@ -23,7 +23,10 @@ contract MarginCalculator {
 
     address public addressBook;
 
+    /// @dev decimals used by strike price and oracle price
     uint256 internal constant BASE = 8;
+
+    /// @dev FixedPoint 0
     FPI.FixedPointInt internal ZERO = FPI.fromScaledUint(0, BASE);
 
     constructor(address _addressBook) public {
--- a/contracts/OtokenFactory.sol
+++ b/contracts/OtokenFactory.sol
@@ -25,6 +25,9 @@ contract OtokenFactory is OtokenSpawner {
     /// @dev mapping from parameters hash to its deployed address
     mapping(bytes32 => address) private idToAddress;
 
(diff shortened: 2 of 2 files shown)


contracts/MarginCalculator.sol
```diff
@@ -23,7 +23,10 @@ contract MarginCalculator {
 
     address public addressBook;
 
+    /// @dev decimals used by strike price and oracle price
     uint256 internal constant BASE = 8;
+
+    /// @dev FixedPoint 0
     FPI.FixedPointInt internal ZERO = FPI.fromScaledUint(0, BASE);
```
aparnakr, line 30: Is this also a constant?

contracts/OtokenFactory.sol
```diff
@@ -25,6 +25,9 @@ contract OtokenFactory is OtokenSpawner {
     /// @dev mapping from parameters hash to its deployed address
     mapping(bytes32 => address) private idToAddress;
 
+    /// @dev max expiry that BokkyPooBahsDateTimeLibrary can handle. (2345/12/31)
+    uint256 private MAX_EXPIRY = 11865398400;
```
aparnakr, line 29: is this a constant?

aparnakr (approved):

contracts/MarginCalculator.sol
antoncoding: it should be but we can't declare struct as constant in solidity.

haythem96 (approved):

aparnakr (approved):

PR #288, opened 2020-11-17
[L03] Operations are not explicitly casted

antoncoding:
# Task: Fix L03

* Explicitly cast number to bytes in `OtokenSpawner`
* Explicitly cast decimals(uint8) to uint256

### Todos
- [x] Unit test 100% coverage

### DIFF
--- a/contracts/MarginCalculator.sol
+++ b/contracts/MarginCalculator.sol
@@ -87,7 +87,7 @@ contract MarginCalculator {
 
         FPI.FixedPointInt memory collateralAmount = ZERO;
         if (hasCollateral) {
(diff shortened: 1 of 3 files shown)


haythem96 (approved):
LGTM! just to make sure, there are no other explicitly casting needed in the codebase :D ?

antoncoding:
> LGTM! just to make sure, there are no other explicitly casting needed in the codebase :D ?

I searched for all uint type which is not `uint256`, and also fix the bytes cased mentioned by oz, but not sure if there's anything else.

aparnakr (commented):
Are there other places in the code base where this is still missing? Do we use uint8 anywhere? What about in margin calculator where we convert from int256 and fixed point and such?

aparnakr (approved):

PR #292, opened 2020-11-18
[L01] Disputer can dispute a price before a pricer has submitted a price

haythem96:
# [L01] Disputer can dispute a price before a pricer has submitted a price

## High Level Description

This PR update the dispute price function to:
- allow disputing only when the price of that certain asset at a specific time is already submitted by the pricer.
- update price submission timestamp to be equal dispute timestamp.

### Code

- [x] Unit test 100% coverage

### Documentation

### DIFF
--- a/contracts/Oracle.sol
+++ b/contracts/Oracle.sol
@@ -221,8 +221,12 @@ contract Oracle is Ownable {
         require(!isDisputePeriodOver(_asset, _expiryTimestamp), "Oracle: dispute period over");
 
         Price storage priceToUpdate = storedPrice[_asset][_expiryTimestamp];
+
+        require(priceToUpdate.timestamp != 0, "Oracle: price to dispute does not exist");
+
         uint256 oldPrice = priceToUpdate.price;
         priceToUpdate.price = _price;
+        priceToUpdate.timestamp = now;
 
         emit ExpiryPriceDisputed(_asset, _expiryTimestamp, oldPrice, _price, now);
     }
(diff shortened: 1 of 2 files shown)


contracts/Oracle.sol
```diff
@@ -221,8 +221,12 @@ contract Oracle is Ownable {
         require(!isDisputePeriodOver(_asset, _expiryTimestamp), "Oracle: dispute period over");
 
         Price storage priceToUpdate = storedPrice[_asset][_expiryTimestamp];
+
+        require(priceToUpdate.timestamp != 0, "Oracle: price to dispute does not exist");
+
         uint256 oldPrice = priceToUpdate.price;
         priceToUpdate.price = _price;
+        priceToUpdate.timestamp = now;
```
aleone, line 229: This has the effect of extending the dispute period each time a price is disputed, correct?

I am fine with that, but wanted to highlight it.

aleone (approved):
How will we deal with chainlink never providing a price?  Will we just pause the contracts and change the pricer?

contracts/Oracle.sol
haythem96: Oh I forgot about that, was wondering why we didn't already include it.
Should we remove it, or add another field to indicate the time of dispute ?

haythem96:
> How will we deal with chainlink never providing a price? Will we just pause the contracts and change the pricer?

Yeah I think we can easily change pricer, worst case scenario is just changing the pricer address to an EOA (wallet) and push price manually.

contracts/Oracle.sol
haythem96: Thought @aparnakr @antoncoding ?

haythem96: Going to remove the timestamp update, and keep the timestamp relate only to the first time where price pushed.

antoncoding (approved):

aparnakr (approved):

PR #297, opened 2020-11-18
[L09] oToken symbols might not be unique

haythem96:
# [L09] oToken symbols might not be unique

## High Level Description

This PR add the collateral asset in the Otoken symbol.

e.g:
before: oETHUSDC-05SEP20-200P
now: oETHUSDC/USDC-05SEP20-200P

### Code

- [x] Unit test 100% coverage

### Documentation

### DIFF
--- a/contracts/Otoken.sol
+++ b/contracts/Otoken.sol
@@ -139,12 +139,13 @@ contract Otoken is ERC20Initializable {
             )
         );
 
-        // concatenated symbol string: oETHUSDCUSDC-05SEP20-200P
+        // concatenated symbol string: oETHUSDC/USDC-05SEP20-200P
         tokenSymbol = string(
             abi.encodePacked(
                 "o",
                 underlying,
                 strike,
+                "/",
                 collateral,
                 "-",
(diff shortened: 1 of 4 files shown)


contracts/Otoken.sol
```diff
@@ -139,12 +139,13 @@ contract Otoken is ERC20Initializable {
             )
         );
 
-        // concatenated symbol string: oETHUSDC-05SEP20-200P
+        // concatenated symbol string: oETHUSDCUSDC-05SEP20-200P
```
aparnakr, line 142: did we also want to have the name use / instead of -? :D

aparnakr: @mshrieve @aleone

aparnakr (approved):

contracts/Otoken.sol
haythem96: Yes will add it.

haythem96: I think the suggestion was to add the `/` before the collateral asset.

antoncoding (approved):

PR #298, opened 2020-11-18
[N01] Functions default to a particular case

haythem96:
# [N01] Functions default to a particular case

## High Level Description

This PR add the specific condition on when to:
- execute the Call action

### Code

- [x] Unit test 100% coverage

### Documentation

### DIFF
--- a/contracts/Controller.sol
+++ b/contracts/Controller.sol
@@ -503,8 +503,7 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
                 _redeem(Actions._parseRedeemArgs(action));
             } else if (actionType == Actions.ActionType.SettleVault) {
                 _settleVault(Actions._parseSettleVaultArgs(action));
-            } else {
-                // actionType == Actions.ActionType.Call
+            } else if (actionType == Actions.ActionType.Call) {
                 ethLeft = _call(Actions._parseCallArgs(action), ethLeft);
             }
         }
--- a/contracts/Otoken.sol
+++ b/contracts/Otoken.sol
@@ -256,7 +256,7 @@ contract Otoken is ERC20Initializable {
             return ("OCT", "October");
         } else if (_month == 11) {
             return ("NOV", "November");
-        } else {
+        } else if (_month == 12) {
             return ("DEC", "December");
         }
     }
--- a/contracts/Otoken.sol
+++ b/contracts/Otoken.sol
@@ -256,7 +256,7 @@ contract Otoken is ERC20Initializable {
             return ("OCT", "October");
         } else if (_month == 11) {
             return ("NOV", "November");
-        } else if (_month == 12) {
+        } else {
             return ("DEC", "December");
         }
     }


haythem96:
Added only a specific else statement for the ` _runActions()`.

aparnakr (commented):
What about the default on month?

haythem96:
> What about the default on month?

I don't think it is needed, also if we add an if statement for month of December, we need a condition to default for if the month is > 12. IMO better to just keep it as it is.

antoncoding (approved):
I don't like this PR because it will bring down our coverage ratio 😢 . But agree that having a default case is kind of dangerous

aparnakr:
> > What about the default on month?
> 
> I don't think it is needed, also if we add an if statement for month of December, we need a condition to default for if the month is > 12. IMO better to just keep it as it is.

why don't we just require the month is 1-12? Having defaults as a code pattern seems to lead to tricky issues. 

Also are there any other places in the code that have defaults?

haythem96:
> > > What about the default on month?
> > 
> > 
> > I don't think it is needed, also if we add an if statement for month of December, we need a condition to default for if the month is > 12. IMO better to just keep it as it is.
> 
> why don't we just require the month is 1-12? Having defaults as a code pattern seems to lead to tricky issues.
> 
> Also are there any other places in the code that have defaults?

There are no other places with default condition.
I don't think we need to restrict the month here, worst case scenario someone will send month == 13 and it will be set to December.

aparnakr:
Eg. https://github.com/opynfinance/GammaProtocol/blob/fix/N01/contracts/libs/MarginVault.sol#L54 Do we want to be explicit here as well?

aparnakr:
> > > > What about the default on month?
> > > 
> > > 
> > > I don't think it is needed, also if we add an if statement for month of December, we need a condition to default for if the month is > 12. IMO better to just keep it as it is.
> > 
> > 
> > why don't we just require the month is 1-12? Having defaults as a code pattern seems to lead to tricky issues.
> > Also are there any other places in the code that have defaults?
> 
> There are no other places with default condition.
> 
> I don't think we need to restrict the month here, worst case scenario someone will send month == 13 and it will be set to December.

For this example it is fine, but as a general coding pattern do we want to try to be explicit as opposed to have defaults? It seems like in other situations it could lead to more unexpected behavior if we are less explicit?

haythem96:
> Eg. https://github.com/opynfinance/GammaProtocol/blob/fix/N01/contracts/libs/MarginVault.sol#L54 Do we want to be explicit here as well?

I updated those in Certora PR https://github.com/opynfinance/GammaProtocol/pull/271/files
We should merge it to master.

aparnakr:
> > Eg. https://github.com/opynfinance/GammaProtocol/blob/fix/N01/contracts/libs/MarginVault.sol#L54 Do we want to be explicit here as well?
> 
> I updated those in Certora PR https://github.com/opynfinance/GammaProtocol/pull/271/files
> We should merge it to master.
 oh, can you make a PR to master from that? The certora specs branch has a lot else going on so idk if we will merge that!

aparnakr (approved):

PR #301, opened 2020-11-19
 [C01] An attacker can steal all the collateral assets

aparnakr:
# Task: Feature Name:
This PR fixed the Controller.sol contract
## High Level Description

Specific Changes
Fix C01 by adding a check in the _redeem function to make sure only valid otokens can be redeemed. 

### Code

- [x] Unit test 100% coverage

### Documentation

### DIFF
--- a/contracts/Controller.sol
+++ b/contracts/Controller.sol
@@ -717,6 +717,8 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
 
         require(isSettlementAllowed(_args.otoken), "Controller: asset prices not finalized yet");
 
+        require(whitelist.isWhitelistedOtoken(_args.otoken), "Controller: otoken is not whitelisted to be redeemed");
+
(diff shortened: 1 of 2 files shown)


haythem96 (commented):
I think we can move the require statement to the beginning of the function.
Also will be great to add unit test using a non-whitelisted Otoken.

contracts/Controller.sol
```diff
@@ -717,6 +717,8 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
 
         require(isSettlementAllowed(_args.otoken), "Controller: asset prices not finalized yet");
 
+        require(whitelist.isWhitelistedOtoken(_args.otoken), "Controller: otoken is not whitelisted to be redeemed");
```
antoncoding, line 720: Add test for this line!

haythem96 (approved):

antoncoding (approved):
Looks good

PR #305, opened 2020-11-19
[M04]: Use of transfer might render ETH impossible to withdraw

aparnakr:
# Use of transfer might render ETH impossible to withdraw

## High Level Description
Changes PayableProxyController.sol

Specific Changes
Instead of using .transfer, now using .sendValue() for sending ETH

### Code

- [x] Unit test 100% coverage

### Documentation

### DIFF
--- a/contracts/external/proxies/PayableProxyController.sol
+++ b/contracts/external/proxies/PayableProxyController.sol
@@ -10,6 +10,7 @@ import {SafeERC20} from "../../packages/oz/SafeERC20.sol";
 import {ERC20Interface} from "../../interfaces/ERC20Interface.sol";
 import {Actions} from "../../libs/Actions.sol";
 import {Controller} from "../../Controller.sol";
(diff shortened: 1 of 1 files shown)


contracts/external/proxies/PayableProxyController.sol
```diff
@@ -70,7 +71,7 @@ contract PayableProxyController is ReentrancyGuard {
             require(sendEthTo != address(0), "PayableProxyController: cannot send ETH to address zero");
 
             weth.withdraw(remainingWeth);
-            sendEthTo.transfer(remainingWeth);
+            Address.sendValue(sendEthTo, remainingWeth);
```
haythem96, line 74: I think will be better to add `using Address for address` in the beginning, and then make the call through `sendEthTo.sendValue(remainingWeth)`

antoncoding (approved):
Agree with Haythem's suggestion ;) but good with this as well ;)

haythem96 (approved):

PR #306, opened 2020-11-20
Remove USDC pricer and integrate stable prices into Oracle module

haythem96:
# Remove USDC pricer and integrate stable prices into Oracle module

## High Level Description

This PR remove the USDC pricer and add a mapping to store stable prices for certain assets.
The stable price can be added/removed only by the Oracle module owner.

Note: any pricer call that rely on a stable asset price will fail if that price is not available in the Oracle module.
e.g: for cUSDC pricer, the stable price of USDC should be pushed to the Oracle module otherwise the cUSDC pricer call at expiry will fail.

### Code

- [x] Unit test 100% coverage

### Documentation


contracts/Oracle.sol
```diff
@@ -205,6 +222,18 @@ contract Oracle is Ownable {
...
+     * @notice set stable asset price
+     * @dev price should be scaled by 1e8
+     * @param _asset asset address
+     * @param _price price
+     */
+    function setStablePrice(address _asset, uint256 _price) external onlyOwner {
+        stablePrice[_asset] = _price;
+
+        emit StablePriceUpdated(_asset, _price);
+    }
```
aparnakr, line 235: maybe add in the comments / documentation that this will only be used for USDC / what other assets? That way its clear to other developers who read our documentation :D

aparnakr (approved):

antoncoding:
> for cUSDC pricer, the stable price of USDC should be pushed to the Oracle module otherwise the cUSDC pricer call at expiry will fail.

This is referring to the issue solved by this PR, right?

haythem96:
> > for cUSDC pricer, the stable price of USDC should be pushed to the Oracle module otherwise the cUSDC pricer call at expiry will fail.
> 
> This is referring to the issue solved by this PR, right?

Yeah.

antoncoding (approved):
CLEAN 😎

PR #307, opened 2020-11-21
Refactor Call action

haythem96:
# Refactor Call action

## High Level Description

This PR remove the following fields from the Call action args:
- owner: can be added in the transaction `data` when needed by a Callee
- vaultId: can be added in transaction `data` when needed by a Callee
- msgValue: not needed anymore as we removed ETH from `Controller.sol`

### Code

- [x] Unit test 100% coverage

### Documentation


haythem96:
Review/merge after H01

antoncoding (approved):
I think we should rename this PR to H01 fix!

aparnakr (approved):

PR #308, opened 2020-11-21
[M02] - ERC20 compliant assets may not be used

aleone:
# Task: Feature Name: [M02] - ERC20 compliant assets may not be used

## High Level Description
require statement to ensure that decimals are within range
require statement to ensure that symbol is not ""
these require statements cause a non-elegant revert if symbols/decimals are not defined which prevents whitelisting of not acceptable erc-20s

unit tests updated to handle branches of symbol/decimal logic

Few ways to handle fetching the symbol/decimal, welcome suggestions if I can improve the implementation.

Initial implementation was calling ERC20Interface(_asset).decimals in the require statement, but switched to declaring variables for decimals and symbol.

I'd think I can also define the asset and call asset.decimals() and asset.symbol() in the require.

### Code

- [x] Unit test 100% coverage

### Documentation

### DIFF
--- a/contracts/Whitelist.sol
+++ b/contracts/Whitelist.sol
@@ -3,6 +3,7 @@ pragma solidity 0.6.10;
 
 import "./interfaces/AddressBookInterface.sol";
 import "./packages/oz/Ownable.sol";
+import {ERC20Interface} from "./interfaces/ERC20Interface.sol";
 
 /**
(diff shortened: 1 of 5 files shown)


aparnakr (approved):
Looks great! <3

haythem96:
Will close this one as we are not applying the fix.

PR #309, opened 2020-11-21
Fix chainlink pricer

haythem96:
# Fix chainlink pricer

## High Level Description

This PR update the chainlink pricer to include a `bot` address that have the privilege to call the `setExpiryPriceInOracle` to push price to Oracle module. 

### Code

- [x] Unit test 100% coverage

### Documentation


antoncoding (approved):
I think we should call the role something else than `bot`. Maybe admin is better

aparnakr:
Why do we need an admin/ bot to be the only ones to be able to call the setExpiryORacle price?

aparnakr (approved):

antoncoding:
> Why do we need an admin/ bot to be the only ones to be able to call the setExpiryORacle price?

Because of how Chainlink work, our (loop) id check can run out of gas. The current fix is to use a privileged address to specify which `roundId` is valid for the expiry.

PR #310, opened 2020-11-23
Fix Compound pricer

haythem96:
# Fix Compound pricer

## High Level Description

This PR refactor the `CompoundPricer.sol` to use the Oracle to get the underlying asset price instead of calling the underlying asset pricer directly, as the Oracle module check if the asset have stable price, then it will return the price, otherwise will forward the call to the asset pricer. This fix the case where the underlying is USDC or ETH for cUSDC or cETH.

This PR also modify the `isLockingPeriodOver()` and `isDisputePeriodOver()` to check if the asset we are checking have a stable price or not, in case the asset have a stable price (e.g USDC) then this asset have no pricer, no locking period and no dispute period, so it always return locking period and dispute period as over. If the asset is not stable the asset, then the same normal checks execute to check if the asset pricer locking or dispute period is over or not.

### Code

- [x] Unit test 100% coverage

### Documentation


contracts/Oracle.sol
```diff
@@ -159,15 +165,22 @@ contract Oracle is Ownable {
      * @return True if dispute period is over, False if not
      */
     function isDisputePeriodOver(address _asset, uint256 _expiryTimestamp) public view returns (bool) {
-        // check if the pricer has a price for this expiry timestamp
-        Price memory price = storedPrice[_asset][_expiryTimestamp];
-        if (price.timestamp == 0) {
-            return false;
+        uint256 price = stablePrice[_asset];
```
aparnakr, line 168: this is good, just wondering, can we accidentally end up with having an asset with a pricer and have it as a stableAsset?

aparnakr: do we need a check when setting the pricer / setting it as stable?

haythem96: mm yeah, e.g if we have a pricer for WETH and set accidentally a stable price for WETH, then we will always return WETH price + no locking period + no dispute period.

haythem96: Good idea, I will add in another PR a check for that.

test/unit-tests/compoundPricer.test.ts
```diff
@@ -111,7 +106,8 @@ contract('CompoundPricer', ([owner, random]) => {
     const usdPrice = createScaledNumber(1)
     const exchangeRate = new BigNumber('211619877757422')
     before('mock data in chainlink pricer and cToken', async () => {
-      await usdcPricer.setPrice(usdPrice)
+      // await usdcPricer.setPrice(usdPrice)
```
aparnakr, line 109:
```suggestion
```

test/unit-tests/compoundPricer.test.ts
```diff
@@ -125,14 +121,15 @@ contract('CompoundPricer', ([owner, random]) => {
...
       const cTokenPrice = await cusdcPricer.getPrice()
       const expectedResult = await underlyingPriceToCtokenPrice(new BigNumber(newPrice), exchangeRate, usdc)
       assert.equal(cTokenPrice.toString(), expectedResult.toString())
     })
-    it('should revert if price is lower than 0', async () => {
-      await usdcPricer.setPrice('0')
-      await expectRevert(cusdcPricer.getPrice(), 'CompoundPricer: underlying price is 0')
+    it('should revert if price is 0', async () => {
+      await oracle.setStablePrice(usdc.address, '0')
+      // await usdcPricer.setPrice('0')
```
aparnakr, line 131:
```suggestion
```

aparnakr (approved):

aparnakr (approved):

aleone (approved):

contracts/Oracle.sol
```diff
@@ -146,10 +146,16 @@ contract Oracle is Ownable {
      * @return True if locking period is over, False if not
      */
     function isLockingPeriodOver(address _asset, uint256 _expiryTimestamp) public view returns (bool) {
-        address pricer = assetPricer[_asset];
-        uint256 lockingPeriod = pricerLockingPeriod[pricer];
+        uint256 price = stablePrice[_asset];
+
+        if (price == 0) {
```
antoncoding, line 151: I think it's cleaner to use a (uint256 => bool) mapping `isStableAsset` to check the type of asset, instead of price == 0.

Or add a function `isStableAsset()` for this repeating check.

aleone (approved):
Looks good.  Agreed on Aparna's comment re checking to ensure that there can not be a pricer as well as a stable price for an asset.

Looks like some MarginCalculator unit tests depend on ctoken logic.  Also saw the ci failed, think because of these failed tests.

contracts/mocks/MockOracle.sol
```diff
@@ -65,7 +75,15 @@ contract MockOracle {
     }
 
     function getExpiryPrice(address _asset, uint256 _expiryTimestamp) external view returns (uint256, bool) {
-        return (storedPrice[_asset][_expiryTimestamp], isFinalized[_asset][_expiryTimestamp]);
+        uint256 price = stablePrice[_asset];
+        bool _isFinalized = true;
+
+        if (price == 0) {
+            price = storedPrice[_asset][_expiryTimestamp];
+            _isFinalized = isFinalized[_asset][_expiryTimestamp];
```
antoncoding, line 83: isn't using isFinalized from the storage (`isFinalized[_asset][_expiryTimestamp]; `) makes it something that need to be poked on?

contracts/mocks/MockOracle.sol
```diff
@@ -31,7 +33,15 @@ contract MockOracle {
...
-        return realTimePrice[_asset];
+        uint256 price = stablePrice[_asset];
+
+        if (price == 0) {
+            require(assetPricer[_asset] != address(0), "Oracle: Pricer for this asset not set");
+
+            price = OpynPricerInterface(assetPricer[_asset]).getPrice();
+        }
+
+        return price;
```
antoncoding, line 44: I don't know why do we need to update the mock contract. The mock contract is supposed to be easily return the mocked value, right? Why using calls to external contract instead of just mock the value?

antoncoding (changes requested):
I think we should leave the mock contract as it is, adding additional logic in the mock contract doesn't add any value, because the point of using them in unit tests is to remove the complicated setup process and easily mock data. (In this case, the stable price setup.) So ideally the developer of MarginCalculator doesn't need to know how to setup an Oracle, he/she just need to know what should his modules behave when the returned values are different. I think that's the whole point of using Mock contracts.

And you shouldn't need to update other contracts' unit tests (in this case, the MarginCalculator). We should only worry about the integration tests when we have this kind of update.

antoncoding (approved):

aleone:
image
Should these be named the same?

contracts/mocks/MockOracle.sol
haythem96: Oh yeah, I updated the mock contract to add the stable price feature, cause it didn't exist before, but yeah we can still not call the pricer and return realTimePrice. But I got different errors so ended up having mock similar to the original contract

contracts/Oracle.sol
haythem96: I thought about it, IMO feel it is just more code, can still use `0` as non-existent as we will not have a stable price equal to 0

haythem96:
> image
> Should these be named the same?

It shouldn't be a problem but I will change the name to not get confused.

contracts/mocks/MockOracle.sol
haythem96: The Compound pricer unit tests use MockOracle file, I think if we don't add the pricer in the MockOracle, then the case of cETH in compound pricer tests does not make sense as we are not using the underlying pricer for testing (WETH pricer)

test/unit-tests/compoundPricer.test.ts
```diff
@@ -125,15 +125,11 @@ contract('CompoundPricer', ([owner, random]) => {
...
-      await usdcPricer.setPrice(newPrice)
+      await oracle.setStablePrice(usdc.address, newPrice)
       const cTokenPrice = await cusdcPricer.getPrice()
       const expectedResult = await underlyingPriceToCtokenPrice(new BigNumber(newPrice), exchangeRate, usdc)
       assert.equal(cTokenPrice.toString(), expectedResult.toString())
     })
-    it('should revert if price is lower than 0', async () => {
-      await usdcPricer.setPrice('0')
-      await expectRevert(cusdcPricer.getPrice(), 'CompoundPricer: underlying price is 0')
-    })
```
haythem96, line 136: Removed this here cause this testing when USDC price = 0?!
In the Oracle mock this will not fail, using the real Oracle contract this will fail with reason of "could not find Pricer"

haythem96:
> > image
> > Should these be named the same?
> 
> It shouldn't be a problem but I will change the name to not get confused.

If this is not making any confusion to other people, I will keep it as it is more efficient to use un-needed declared variables :D

test/unit-tests/compoundPricer.test.ts
aparnakr: can you change the test to make sure it reverts?

aparnakr (approved):

contracts/Oracle.sol
antoncoding: How about a function called `isStableAsset()` and in that function check if `stablePrice != 0`?

contracts/Oracle.sol
haythem96: #311

aparnakr (approved):

test/unit-tests/compoundPricer.test.ts
aparnakr: don't delete this test!

haythem96: The Compound pricer us e a mock, and the mock does not use a real underlying pricer, changing this will require alot of refactoring I think.
Any missing use cases should be added to the integration tests.

Also this test does not make sense after merging #311 as an asset should have a stable price > 0 or a pricer contract.

cc @antoncoding

PR #312, opened 2020-11-24
Compound tokens wrapper

haythem96:
# Compound tokens wrapper

## High Level Description
-Operator that can wrap and unwrap erc-20 based cTokens for users before (wrap) or after (unwrap) running Action args via the Controller operate function.
-Returns excess underlying asset to the user that was sent but not used

### Code

- [x] Unit test 100% coverage

### Documentation


contracts/external/proxies/CERC20Proxy.sol
```diff
@@ -0,0 +1,90 @@
...
+                require(
+                    (msg.sender == action.owner) || (controller.isOperator(action.owner, msg.sender)),
+                    "CERC20Proxy: msg.sender is not owner or operator "
+                );
+            }
+
+            // overwrite the deposit amount by the exact amount minted
+            if (action.actionType == Actions.ActionType.DepositCollateral && action.amount == 0) {
+                // should we only update if the action.second address == address(this)?
+                _actions[i].amount = cTokenBalance;
```
aparnakr, line 76: why not just overwrite it always?

contracts/external/proxies/CERC20Proxy.sol
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        controller.operate(_actions);
+
+        // unwrap and withdraw cTokens that have been added to contract via operate function
+        uint256 cTokenBalanceAfter = cToken.balanceOf(address(this));
+        if (cTokenBalanceAfter > 0) {
+            require(cToken.redeem(cTokenBalanceAfter) == 0, "CTokenPricer: Redeem Failed");
+            uint256 underlyingBalance = underlying.balanceOf(address(this));
+            underlying.safeTransfer(msg.sender, underlyingBalance);
```
aparnakr, line 87: we should make it clear in the docs that the end user should pass in an action which transfers underlying to this address for redeeming underlying to work

contracts/external/proxies/CERC20Proxy.sol
```diff
@@ -0,0 +1,90 @@
...
+            // overwrite the deposit amount by the exact amount minted
+            if (action.actionType == Actions.ActionType.DepositCollateral && action.amount == 0) {
+                // should we only update if the action.second address == address(this)?
+                _actions[i].amount = cTokenBalance;
+            }
+        }
+
+        controller.operate(_actions);
+
+        // unwrap and withdraw cTokens that have been added to contract via operate function
```
aparnakr, line 82: we should make it clear in the docs that the end user should pass in an action which transfers underlying to this address for redeeming underlying to work. We should also make it clear in the docs of who the "from" field should take money from, "to" field to send money to should be

aparnakr (approved):

contracts/external/proxies/CERC20Proxy.sol
aleone: It lets people deposit partial cToken and partial underlying asset in the same tx. They can have two DepositCollateral actions one with cTokens and amt !=0 and one with cTokens but amt =0.

contracts/external/proxies/CERC20Proxy.sol
aleone: Yeah the action args for this will need some clear documentation, anything to wrap/unwrap has to go to/from this address.

contracts/external/proxies/CERC20Proxy.sol
aleone: agreed, similar to above.  Args are a bit confusing through the operator when compared to normal args.

contracts/external/proxies/CERC20Proxy.sol
antoncoding: We should remove the comment `// should we only update if the action.second address == address(this)?` 
as user shouldn't put 0 if they don't intend to do this.

contracts/external/proxies/CETHProxy.sol
```diff
@@ -0,0 +1,90 @@
...
+     * @param _underlying underlying asset
+     * @param _cToken the cToken to mint
+     */
+    function operate(
+        Actions.ActionArgs[] memory _actions,
+        address _underlying,
+        address _cToken
+    ) external payable nonReentrant {
+        require(_underlying == address(0), "CETHWrapUnwrap: ETH address other than address(0) specified");
+        require(_cToken == cethAddress, "CETHWrapUnwrap: Wrong cToken address specified for ETH");
```
antoncoding, line 50:
```suggestion
    function operate(
        Actions.ActionArgs[] memory _actions,
        address _underlying
    ) external payable nonReentrant {
        require(_underlying == address(0), "CETHWrapUnwrap: ETH address other than address(0) specified");
```
In case of cETH, I don't think we need to pass the ctoken type again, can just use predefined cethAddress

contracts/external/proxies/CETHProxy.sol
```diff
@@ -0,0 +1,90 @@
...
+
+        CETHInterface cToken = CETHInterface(_cToken);
```
antoncoding, line 52:
```suggestion
        CETHInterface cToken = CETHInterface(cethAddress);
```

antoncoding (approved):
Should cleanup a bit more before merging

PR #314, opened 2020-11-29
[Certora 13]: Ensure that the margin pool cannot transfer assets to/from itself

aparnakr:
# [Certora 13]: Ensure that the margin pool cannot transfer assets to/from itself

## High Level Description
Checks that the margin pool cannot transfer assets to / from itself. 

Specific Changes
Added a require condition in transferToPool and transferToUser functions in the marginPool.sol file. 

### Code

- [x] Unit test 100% coverage

### Documentation


antoncoding (approved):

contracts/MarginPool.sol
```diff
@@ -77,7 +77,7 @@ contract MarginPool is Ownable {
         uint256 _amount
     ) public onlyController {
         require(_amount > 0, "MarginPool: transferToPool amount is equal to 0");
-
+        require(_user != address(this), "MarginPool: cannot transfer assets from oneself");
```
haythem96, line 80: This check is not needed as the `safeTransferFrom` will fail if `_user == address(this)`.

haythem96 (changes requested):
This looks great! I think we just need to remove the check in the `TransferToPool()` as if the `_user == address(pool)`, the `msg.sender` is always the pool address, then the `safeTransferFrom` will fail with `ERC20: transfer amount exceeds allowance`

haythem96 (approved):

PR #330, opened 2021-01-03
Otoken.sol Minor Gas Saving

AlphaSerpentis:
# Task: Otoken.sol Minor Potential Gas Saving

## High Level Description

- require() added to mint if the amount is zero to prevent wasted gas.
- burn function has an if statement check that returns nothing if the amount is zero. **Unsure if to put a require() in this spot in the event some other operation is running alongside.**
- Removed setting variable `uint256 trailingZeroes = 0;` and replaced for `uint256 trailingZeroes;` for gas saving.

### Code

- [ ] Unit test 100% coverage

### Documentation


antoncoding:
Thanks for opening this! for the 3 updates:

1. For the require statement in mint, would be nice to include that in the unit test as well. 
2. For the burn function, I prefer to use a require statement so it's more consistent throughout the codebase. (we never did early `return`). 
3. The `= 0` initialization is a good catch 😄! We did want to remove those for gas saving but I guess we missed this

haythem96:
Thanks for making this PR @AlphaSerpentis 🔥

But I guess I'm not sure if this will reduce required gas, wouldn't the new require statements increase the gas needed when calling the `operate()` in `Controller.sol` ? I think this implementation will save gas for users that are mistakenly minting or burning `0` amount, but If a user is minting 1 Otoken, the required gas will be higher, right ?
Please let me know if I'm missing something :D

AlphaSerpentis:
Hello,

I ran some tests on Remix IDE with optimizer @ 200 runs. Each require statement added about 23 wei. The gas cost increase for execution is indeed a bit higher as a result, but for anyone building on top of the Gamma Protocol, personally, I think it'd be better to prevent them from wasting gas as a result.

If this isn't the case, I suppose at least doing the `= 0` initialization fix.

haythem96:
> Hello,
> 
> I ran some tests on Remix IDE with optimizer @ 200 runs. Each require statement added about 23 wei. The gas cost increase for execution is indeed a bit higher as a result, but for anyone building on top of the Gamma Protocol, personally, I think it'd be better to prevent them from wasting gas as a result.
> 
> If this isn't the case, I suppose at least doing the `= 0` initialization fix.

Awesome. Yeah I think no one should mint or burn 0 amount accidentally, and if it happens I think it is better to keep the current implementation as it is cheaper for users who execute transactions.

The `=0` can be applied, even if it is not a fix for a specific issues as it is just another way of writing code, but please feel free to make another PR with that fix :)

AlphaSerpentis:
Should I edit this PR instead, or would you prefer me to close it and open a new PR?

PR #342, opened 2021-01-08
Refactor MarginCalculator

haythem96:
# Refactor MarginCalculator

## High Level Description

This PR refactor the `MarginCalculator.sol` contract to reduce gas consumption.
- Removed the `addressBook` variable. The new implementation will set the `oracle` address at the constructor level instead of making an external call to the `AddressBook` whenever the Oracle is needed.
- Added `getVaultDetails()` function that will get the vault details as a first step in `getExcessCollateral()` and store those details in the `VaultDetails` struct. The old implementation make multiple repeated external calls to get Otoken details (e.g strike asset, underlying asset, isPut, strike price..etc), this refactoring will store all those data locally and pass the struct to the internal functions when needed.
- `getExpiredPayoutRate()` refactor https://github.com/opynfinance/GammaProtocol/pull/342/files#diff-841eb3a84b542db242aa48c8ca2e67a3e8ef17861081197d0aebb1738992b592R69
- `_getExpiredCashValue()` refactor https://github.com/opynfinance/GammaProtocol/pull/342/files#diff-841eb3a84b542db242aa48c8ca2e67a3e8ef17861081197d0aebb1738992b592R149

**Ignore the changes made in the `specs` folder as they were made automatically by the linter.**

### Code

- [x] Unit test 100% coverage

### Documentation

### DIFF
--- a/contracts/MarginCalculator.sol
+++ b/contracts/MarginCalculator.sol
@@ -29,6 +29,27 @@ contract MarginCalculator {
     /// @dev FixedPoint 0
     FPI.FixedPointInt internal ZERO = FPI.fromScaledUint(0, BASE);
 
+    struct VaultDetails {
+        address shortUnderlyingAsset;
+        address shortStrikeAsset;
+        address shortCollateralAsset;
+        address longUnderlyingAsset;
+        address longStrikeAsset;
+        address longCollateralAsset;
+        uint256 shortStrikePrice;
+        uint256 shortDecimals;
+        uint256 shortExpiryTimestamp;
+        uint256 longStrikePrice;
+        uint256 longDecimals;
+        uint256 longExpiryTimestamp;
+        uint256 collateralDecimals;
+        bool shortType;
+        bool longType;
+        bool hasLong;
+        bool hasShort;
+        bool hasCollateral;
+    }
+
     constructor(address _addressBook) public {
         require(_addressBook != address(0), "MarginCalculator: invalid addressbook");
 
@@ -75,6 +96,8 @@ contract MarginCalculator {
      * if True, collateral can be taken out from the vault, if False, additional collateral needs to be added to vault
      */
     function getExcessCollateral(MarginVault.Vault memory _vault) public view returns (uint256, bool) {
+        // get vault details
+        VaultDetails memory vaultDetails = getVaultDetails(_vault);
         // include all the checks for to ensure the vault is valid
         _checkIsValidVault(_vault);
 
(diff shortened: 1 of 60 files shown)


contracts/MarginCalculator.sol
```diff
@@ -21,18 +20,40 @@ contract MarginCalculator {
...
+        address longCollateralAsset;
+        uint256 shortStrikePrice;
+        uint256 shortExpiryTimestamp;
+        uint256 shortCollateralDecimals;
+        uint256 longStrikePrice;
+        uint256 longExpiryTimestamp;
+        uint256 longCollateralDecimals;
+        uint256 collateralDecimals;
+        bool shortType;
+        bool longType;
```
antoncoding, line 47: I think it's better to call them isShortPut, isLongPut.
IMO it's nontrivial to see a boolean variable that's called "type".

contracts/MarginCalculator.sol
```diff
@@ -302,53 +332,56 @@ contract MarginCalculator {
...
-            long.strikeAsset() == short.strikeAsset() &&
-            long.collateralAsset() == short.collateralAsset() &&
-            long.expiryTimestamp() == short.expiryTimestamp() &&
-            long.isPut() == short.isPut();
+            _vaultDetails.longUnderlyingAsset == _vaultDetails.shortUnderlyingAsset &&
+            _vaultDetails.longStrikeAsset == _vaultDetails.shortStrikeAsset &&
+            _vaultDetails.longCollateralAsset == _vaultDetails.shortCollateralAsset &&
+            _vaultDetails.longExpiryTimestamp == _vaultDetails.shortExpiryTimestamp &&
+            _vaultDetails.longType == _vaultDetails.shortType;
     }
```
antoncoding, line 366: didn't realize how expensive this was until looking at this diff 😮

antoncoding (approved):
Would be nice to exclude those Certora changes out of this PR!

contracts/MarginCalculator.sol
haythem96: yeah external calls are 💸

haythem96:
> Would be nice to exclude those Certora changes out of this PR!

I didn't meant to add those, no idea why the linter changed it in this branch and can't do the same on master 😅

haythem96:
Screen Shot 2021-01-10 at 8 48 03 PM
Screen Shot 2021-01-10 at 8 48 15 PM

contracts/MarginCalculator.sol
```diff
@@ -43,16 +64,17 @@ contract MarginCalculator {
...
     function getExpiredPayoutRate(address _otoken) external view returns (uint256) {
         require(_otoken != address(0), "MarginCalculator: Invalid token address");
+
         OtokenInterface otoken = OtokenInterface(_otoken);
-        require(now > otoken.expiryTimestamp(), "MarginCalculator: Otoken not expired yet");
+        uint256 expiry = otoken.expiryTimestamp();
+
+        require(now > expiry, "MarginCalculator: Otoken not expired yet");
 
         FPI.FixedPointInt memory cashValueInStrike = _getExpiredCashValue(_otoken);
```
antoncoding, line 74: Can we not use overload on `_getExpiredCashValue` and change this to use the new `_getExpiredCashValue`? 
I don't see the old `_getExpiredCashValue` worth existing since it's only used once here.

haythem96: the `_getExpiredCashValue()` is used in `getExpiredPayoutRate()` that only take an Otoken as arg, and does not use a vault, I don't see a way to refactor it using the `VaultDetails` struct.

haythem96: Actually I think I'm wrong, we can use only the new `_getExpiredCashValue` implementation... I'm thinking of a small refactoring in the Controller also that will reduce gas for `_redeem`. Will include the new refactor there.

test/unit-tests/marginCalculator.test.ts
```diff
@@ -150,41 +150,71 @@ contract('MarginCalculator', () => {
...
       const ethPirce = scaleNum(200)
       await oracle.setExpiryPrice(weth.address, closeExpiry, ethPirce)
-      const cashedValue = await calculator.getExpiredCashValue(put.address)
+      const underlying = await put.underlyingAsset()
+      const strike = await put.strikeAsset()
+      const expiryTimestamp = await put.expiryTimestamp()
+      const strikePrice = await put.strikePrice()
+      const isPut = await put.isPut()
+      const cashedValue = await calculator.getExpiredCashValue(underlying, strike, expiryTimestamp, strikePrice, isPut)
       assert.equal(cashedValue.toString(), scaleNum(50))
```
antoncoding, line 159: I think we should have an util function to get these batched info

contracts/MarginCalculator.sol
```diff
@@ -126,35 +135,35 @@ contract MarginCalculator {
...
+    //         otoken.strikeAsset(),
+    //         otoken.expiryTimestamp()
+    //     );
+
+    //     if (otoken.isPut()) {
+    //         return strikePrice.isGreaterThan(underlyingPriceInStrike) ? strikePrice.sub(underlyingPriceInStrike) : ZERO;
+    //     } else {
+    //         return underlyingPriceInStrike.isGreaterThan(strikePrice) ? underlyingPriceInStrike.sub(strikePrice) : ZERO;
+    //     }
+    // }
```
antoncoding, line 166: Delete it!

antoncoding (approved):

test/unit-tests/marginCalculator.test.ts
haythem96: Yup, can do it in another tests cleaning PR.

haythem96:
Screen Shot 2021-01-11 at 3 57 10 PM

contracts/MarginCalculator.sol
```diff
@@ -143,26 +178,39 @@ contract MarginCalculator {
...
-            : OtokenInterface(_vault.longOtokens[0]);
-        bool expired = now > otoken.expiryTimestamp();
-        bool isPut = otoken.isPut();
+    function _getMarginRequired(MarginVault.Vault memory _vault, VaultDetails memory _vaultDetails)
+        internal
+        view
+        returns (FPI.FixedPointInt memory)
+    {
+        FPI.FixedPointInt memory shortAmount = _vaultDetails.hasShort
+            ? FPI.fromScaledUint(_vault.shortAmounts[0], BASE)
```
aparnakr, line 187: why not store _vault.shortAmounts[0] also in vault details? Then you don't need to pass around the vault right? Does that save gas?

contracts/MarginCalculator.sol
```diff
@@ -143,26 +178,39 @@ contract MarginCalculator {
...
+            ? _vaultDetails.shortCollateralAsset
+            : _vaultDetails.longCollateralAsset;
+        address otokenStrikeAsset = _vaultDetails.hasShort
+            ? _vaultDetails.shortStrikeAsset
+            : _vaultDetails.longStrikeAsset;
+        uint256 otokenExpiry = _vaultDetails.hasShort
+            ? _vaultDetails.shortExpiryTimestamp
+            : _vaultDetails.longExpiryTimestamp;
+        bool expired = now > otokenExpiry;
+        bool isPut = _vaultDetails.hasShort ? _vaultDetails.isShortPut : _vaultDetails.isLongPut;
```
aparnakr, line 206: can you check if vaultDetails.hasShort {
// assign shortDetails
} else { 
// assign long details
} 
Does that save more gas than the check every time?

contracts/MarginCalculator.sol
```diff
@@ -21,18 +20,40 @@ contract MarginCalculator {
     using SafeMath for uint256;
     using FPI for FPI.FixedPointInt;
 
-    address public addressBook;
+    /// @dev oracle module
+    OracleInterface public oracle;
```
aparnakr, line 24: does this mean if we change the oracle we have to redeploy margin calculator?

aparnakr (approved):
Looks good to me! This is awesome! :D

aparnakr:
The way I thought about this at a high level was that since all the margin calculator functions are view, then none of the operations change the data in other places, so it doesn't matter if you get all the data first and do the operations later OR if you get the data as and when you need to do the operation :D This is very clever !

contracts/MarginCalculator.sol
haythem96: yeah

contracts/MarginCalculator.sol
haythem96: Oh let me try that in a different branch

contracts/MarginCalculator.sol
haythem96: Yeah I think so, I tried doing that in a diff branch https://github.com/opynfinance/GammaProtocol/pull/345 but broke many tests and couldn't find time to look at that. Maybe if I had time this weekend otherwise we just deploy this as it is for now.

PR #344, opened 2021-01-11
Refactor Otoken minting

haythem96:
# Refactor Otoken minting

## High Level Description

Remove external calls to `AddressBook`.
To apply this change, need to redeploy Factory and new Otoken implementation.
With this change, we need to deploy new Otoken implementation whenever we upgrade Controller.

### Code

- [x] Unit test 100% coverage

### Documentation

### DIFF
--- a/contracts/Otoken.sol
+++ b/contracts/Otoken.sol
@@ -51,7 +51,7 @@ contract Otoken is ERC20Initializable {
      * @param _isPut True if a put option, False if a call option
      */
     function init(
-        address _addressbook,
+        address _addressBook,
         address _underlyingAsset,
         address _strikeAsset,
         address _collateralAsset,
@@ -59,7 +59,7 @@ contract Otoken is ERC20Initializable {
         uint256 _expiryTimestamp,
         bool _isPut
     ) external initializer {
-        controller = AddressBookInterface(_addressbook).getController();
+        controller = AddressBookInterface(_addressBook).getController();
         underlyingAsset = _underlyingAsset;
         strikeAsset = _strikeAsset;
         collateralAsset = _collateralAsset;
--- a/contracts/Otoken.sol
+++ b/contracts/Otoken.sol
@@ -42,7 +42,7 @@ contract Otoken is ERC20Initializable {
 
     /**
      * @notice initialize the oToken
-     * @param _addressbook addressbook module
+     * @param _addressBook addressbook module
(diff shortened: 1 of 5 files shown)


haythem96:
Screen Shot 2021-01-11 at 7 51 42 AM

antoncoding (approved):
From my understanding, this refactor works perfectly because the fact that **we will never update our controller address in `Addressbook`**, which I think make sense but want every body to give it a thought as well.

haythem96:
> From my understanding, this refactor works perfectly because the fact that **we will never update our controller address in `Addressbook`**, which I think make sense but want every body to give it a thought as well.

Ah yeah that's true, I forgot that we don't need to change the Controller address as the Proxy will always forward to the right implementation.

antoncoding:
From our discussion:
Better keep the addressbook in the oToken init function to avoid updating factory contract. We can store the controller address locally during initialization so we can reduce that 1 external call every time mint and burn is called.

haythem96:
> From our discussion:
> Better keep the addressbook in the oToken init function to avoid updating factory contract. We can store the controller address locally during initialization so we can reduce that 1 external call every time mint and burn is called.

Yeah agree, I updated the code.

contracts/Otoken.sol
```diff
@@ -16,8 +16,8 @@ import {AddressBookInterface} from "./interfaces/AddressBookInterface.sol";
 contract Otoken is ERC20Initializable {
     using SafeMath for uint256;
 
-    /// @notice address of the AddressBook module
-    address public addressBook;
+    /// @notice address of the Controller module
+    address public controller;
```
aparnakr, line 20: if we upgrade the controller, is there a way to set this again?

aparnakr (commented):
It looks good! Left a question which we should figure out first!

contracts/Otoken.sol
haythem96: Yeah, when we upgrade the controller, the controller address does not change, it is always the same proxy address, the only address that change is the implementation address which is all handled by the Proxy mechanism.

antoncoding (approved):

contracts/Otoken.sol
aparnakr: can we test upgrading the controller with this?

aparnakr (approved):

PR #348, opened 2021-01-14
Feature/0x callee

aparnakr:
# Feature: Add 0xCallee

## High Level Description
This PR adds a new callee contract which allows for buying and selling from 0x. 

Specific Changes
1. Contracts/Callees/0xCallee.sol is the callee contract that allows for buying and selling. It calls the batchFillOrders method from [0x](https://0x.org/docs/guides/v3-specification#protocol-fees).  
2. test/mainnet-fork/ uses a fork of mainnet data and tests opening a new naked and spread put position.

### Code

- [x] Unit test 100% coverage

### Documentation


contracts/external/callees/Trade0x.sol
```diff
@@ -0,0 +1,167 @@
...
+        for (uint256 i = 0; i < order.length; i++) {
+            // transfer swapped token to sender
+            address makerAsset = decodeERC20Asset(order[i].makerAssetData);
+            uint256 balance = ERC20Interface(makerAsset).balanceOf(address(this));
+            ERC20Interface(makerAsset).safeTransfer(_sender, balance);
+
+            // transfer the taker asset back to the user if the order wasn't fully filled
+            address takerAsset = decodeERC20Asset(order[i].takerAssetData);
+            balance = ERC20Interface(takerAsset).balanceOf(address(this));
+            ERC20Interface(takerAsset).safeTransfer(_sender, balance);
```
antoncoding, line 108: we should probably check if( balance > 0) here? cause users may fill multiple orders with same makerAsset. Otherwise the safeTransfer will fail if we call with amount = 0 (not sure exactly how it's gonna behave)

aparnakr: okay, let me test how it behaves on amount = 0.

contracts/external/callees/Trade0x.sol
```diff
@@ -0,0 +1,161 @@
...
+        require(msg.sender == controller, "sender not controller");
+        _directlyTrade(_sender, _data);
+    }
+
+    function _directlyTrade(address payable _sender, bytes memory _data) internal {
+        (
+            IZeroXExchange.Order[] memory order,
+            uint256[] memory takerAssetFillAmount,
+            bytes[] memory signature,
+            address feePayer
```
haythem96, line 71: Just to make clear here, we can't let the `feePayer` max approve this Contract as any other user can use the same `feePayer` address and pay fee from that wallet, if it have enough WETH available, right ?

aparnakr: Yeah, so if the fee payer is us, then its fine. but yes, otherwise this is an issue and every time the user makes a transaction they have to approve this contract to spend weth.

antoncoding: why can't we check relationship between `_feePayer` and `_sender` in this function? In most cases, this `_feePayer` will be `_sender`:
1. User call Controller directly, `_sender` = user address, and `_feePayer` = user address
2. User call PayableProxy => Controller, `_sender` = payable proxy, `feePayer` should also be the payable proxy. (User send eth to payableProxy => convert ETH to WETH and stay in the payable proxy contract => Being pulled by 0xCallee).

And if _sender != _feePayer, `require Controller.isOperator(_feePayer, _sender) === true`, in case user want to use another operator but still let our callee just pull weth from their wallet.

antoncoding (changes requested):
Can you also add unit test? maybe a simple Mock contract for 0x that pull token from the maker to stimulate the trade.

image

PR #355, opened 2021-01-19
Restrict products

aparnakr:
# Task: Restrict products 

## High Level Description
Only allow creating options products which are fully collateralized which won't require any liquidation mechanism. 

Specific Changes
Added a require statement in whitelist to check that only put option products where the strike == collateral are allowed and call options where underlying == collateral are allowed. 

### Code

- [x] Unit test 100% coverage

### Documentation


antoncoding:
Plz add tests for this line!

haythem96:
This will break the integration tests in `Yield Farming: Naked Put Option closed before expiry flow`, do u want to just ignore those for now by adding `xit`

contracts/Whitelist.sol
```diff
@@ -167,7 +171,7 @@ contract Whitelist is Ownable {
 
     /**
      * @notice allows the owner to whitelist a collateral address
-     * @dev can only be called from the owner address
+     * @dev can only be called from the owner address. Don't whitelist otokens as collateral in the current version.
```
antoncoding, line 174: Why do we want to add this description? otokens are never meant to be a collateral in our system right?

antoncoding (approved):

contracts/Whitelist.sol
haythem96: Yeah I think better to update this comment. Otoken are already collateral in our system, but they are used in the long Otoken array, so maybe make sense to just explain that this functions is meant for collateral assets that are not an Otoken?

haythem96 (approved):

PR #356, opened 2021-01-19
0x callee

haythem96:
# Task: 0x Callee

## High Level Description

This PR add `Trade0x.sol` callee to batch Gamma actions with 0x trading transactions.

### Code

- [ ] Unit test 100% coverage

### Documentation


contracts/external/callees/Trade0x.sol
```diff
@@ -0,0 +1,168 @@
...
+            tx.origin == trader,
+            "Trade0x: funds can only be transferred in from the person sending the transaction"
+        );
+
+        for (uint256 i = 0; i < order.length; i++) {
+            address takerAsset = decodeERC20Asset(order[i].takerAssetData);
+            // transfer takerAsset from trader to this contract
+            ERC20Interface(takerAsset).safeTransferFrom(trader, address(this), takerAssetFillAmount[i]);
+            // approe the 0x ERC20 Proxy to transfer takerAsset from this contract
+            ERC20Interface(takerAsset).safeIncreaseAllowance(assetProxy, takerAssetFillAmount[i]);
```
aparnakr, line 88: one thing to note here is that if the taker asset is weth this will fail. We are not using weth as taker asset though right? So maybe we should document that? @antoncoding

test/e2e/zeroXtrading.ts
```diff
@@ -0,0 +1,448 @@
+// // mainnet fork for 0x trading through 0xCallee
+// import {
+//   PermitCalleeInstance,
+//   Trade0xInstance,
+//   MockERC20Instance,
+//   OtokenInstance,
+//   WETH9Instance,
```
aparnakr, line 7: should we remove this file for now and add it back later once it passes?

contracts/external/callees/Trade0x.sol
haythem96: In case `takerAsset` is WETH, why it will fail ?

aparnakr (approved):
Looks good to me!

contracts/external/callees/Trade0x.sol
aparnakr: Oh I had this issue yesterday where because in the constructor we do a big allowance for WETH, when we call this again, it runs into a safe math overflow issue.

contracts/external/callees/Trade0x.sol
```diff
@@ -0,0 +1,168 @@
...
+            address trader,
+            ZeroXExchangeInterface.Order[] memory order,
+            uint256[] memory takerAssetFillAmount,
+            bytes[] memory signature
+        ) = abi.decode(_data, (address, ZeroXExchangeInterface.Order[], uint256[], bytes[]));
+
+        require(
+            tx.origin == trader,
+            "Trade0x: funds can only be transferred in from the person sending the transaction"
+        );
```
antoncoding, line 81: why not just use
```
address trade = tx.origin
```
Will safe us some data bytes and reduce gas cost

contracts/external/callees/Trade0x.sol
antoncoding: IMO okay to leave it this way since we're unlikely to enable weth / oToken orders in the short term. But we should def add that to comments.

antoncoding, line 88: IMO it's better to do unlimited approval: if (allowance < amount), approve uint256(-1). 
Otherwise we will have 1 additional update to token contract storage, for **each order we take**, which is MUCH more gas. 
(basically imagine gas cost for one approve per order.)

test/e2e/zeroXtrading.ts
antoncoding: remove this.

antoncoding (changes requested):
I don't see any unit test for the actual swapping here. We need at least 1 test with at least 1 order.

image

antoncoding (approved):

antoncoding:
Saw the tests in e2e folder now, we should ignore this file (0xcalle) from or unit test coverage report if we're not gonna write any

haythem96:
> Saw the tests in e2e folder now, we should ignore this file (0xcalle) from or unit test coverage report if we're not gonna write any

I think we keep it, we failed to keep a good dev process for this update as we can make some contracts to mimic 0x trade and increase coverage. So yeah maybe we open an issue for this and also other contributors can work on it.

antoncoding:
> > Saw the tests in e2e folder now, we should ignore this file (0xcalle) from or unit test coverage report if we're not gonna write any
> 
> I think we keep it, we failed to keep a good dev process for this update as we can make some contracts to mimic 0x trade and increase coverage. So yeah maybe we open an issue for this and also other contributors can work on it.

I don't think we'll ever write those if we don't do them now TBH. I'd prefer that we reach consensus on what unit test are necessary and ignore the others less important tests, instead of assuming we will have time to write them all in the future.

For me, saying we will do it in the future is worse then saying we won't do it, cause that mean we're okay about skipping the process that we all agreed on.

PR #359, opened 2021-01-24
Adding 0x callee unit tests

aparnakr:
# Task: Adding 0x callee unit tests

## High Level Description

Unit tests for 0x Callee contract.

### Code

- [ ] Unit test 100% coverage

### Documentation


contracts/mocks/Mock0xExchange.sol
```diff
@@ -29,11 +29,22 @@ contract Mock0xExchange {
...
-    ) external payable returns (IZeroXExchange.FillResults memory fillResults) {
+    ) public payable returns (IZeroXExchange.FillResults memory fillResults) {
         takerAmount = _order.takerAssetAmount;
         makerAmount = _order.makerAssetAmount;
         signature = _signature;
         fillAmount = _takerAssetFillAmount;
         return IZeroXExchange.FillResults(0, 0, 0, 0, 0);
     }
+
+    function batchFillOrders(
```
haythem96, line 40: `batchFillOrders` returns `(LibFillResults.FillResults[] memory fillResults)`, I think we should mock the same returns result and return back an array with the same length as orders.

haythem96:
Can we test minting Otoken (put as an example), sell it, so we can test the permit callee + 0x callee?

contracts/external/callees/Trade0x.sol
```diff
@@ -82,14 +82,14 @@ contract Trade0x is CalleeInterface {
...
-            // for test, will remove it later
-            require(exchange.isValidOrderSignature(order[i], signature[i]), "order signature not valid");
+            // // just for test, will remove later
+            // require(
+            //     ERC20Interface(takerAsset).allowance(trader, address(this)) == takerAssetFillAmount[i],
+            //     "approve not working man!"
+            // );
+
+            // // for test, will remove it later
+            // require(exchange.isValidOrderSignature(order[i], signature[i]), "order signature not valid");
```
antoncoding, line 92: Remove these

aparnakr:
> Can we test minting Otoken (put as an example), sell it, so we can test the permit callee + 0x callee?

to do that we need to deploy 0x locally, if not selling does not actually work. That takes significant work, it is easier to just test it on the mainnet fork

contracts/mocks/MockERC20.sol
```diff
@@ -16,4 +16,13 @@ contract MockERC20 is ERC20Upgradeable {
     function mint(address account, uint256 amount) public {
         _mint(account, amount);
     }
+
+    function transferFrom(
```
haythem96, line 20: Why adding this function ?

PR #368, opened 2021-03-08
Controller modifications for partial collateralization

haythem96:
# Task: Controller modifications for partial collateralization

## High Level Description

This PR add:
- Vault type and vault latest update timestamp storage.
- Add vaultType in open vault action arg and encode it from data field.
- Set vaultLatestUpdate after every successful operate call that alter the vault state.
- Set vault type when opening a vault.
- Update `VaultOpened` event to include vault type.

### Code

- [x] Unit test 100% coverage

### Documentation

### DIFF
--- a/contracts/Controller.sol
+++ b/contracts/Controller.sol
@@ -346,7 +346,10 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
      */
     function operate(Actions.ActionArgs[] memory _actions) external nonReentrant notFullyPaused {
         (bool vaultUpdated, address vaultOwner, uint256 vaultId) = _runActions(_actions);
-        if (vaultUpdated) _verifyFinalState(vaultOwner, vaultId);
+        if (vaultUpdated) {
+            _verifyFinalState(vaultOwner, vaultId);
+            vaultLatestUpdate[vaultOwner][vaultId] = now;
+        }
     }
 
     /**
@@ -530,8 +533,6 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
         (, bool isValidVault) = calculator.getExcessCollateral(_vault);
 
         require(isValidVault, "Controller: invalid final vault state");
-
-        vaultLatestUpdate[_owner][_vaultId] = now;
     }
 
     /**
--- a/contracts/Controller.sol
+++ b/contracts/Controller.sol
@@ -70,7 +70,7 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
     /// @notice emits an event when an account operator is updated for a specific account owner
     event AccountOperatorUpdated(address indexed accountOwner, address indexed operator, bool isSet);
     /// @notice emits an event when a new vault is opened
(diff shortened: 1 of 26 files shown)


haythem96:
Certora rules need some updates in another PR @aparnakr

contracts/Controller.sol
```diff
@@ -60,10 +60,17 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
     /// @dev mapping between an account owner and their approved or unapproved account operators
     mapping(address => mapping(address => bool)) internal operators;
 
+    /******************************************************************** V2.0.0 storage upgrade ******************************************************/
```
aparnakr, line 63: how do other projects do this?

aparnakr: Maybe /*** text ***/ ? https://github.com/compound-finance/compound-protocol/blob/master/contracts/ComptrollerG6.sol#L92 like this?

aparnakr: Not sure stylistically what the best way of separating it is

contracts/libs/Actions.sol
```diff
@@ -82,6 +82,8 @@ library Actions {
         address owner;
         // vault id to create
         uint256 vaultId;
+        // vault type, 0 for spread/max loss and 1 for naked margin vault
+        uint256 vaultType;
```
aparnakr, line 86: are we adding this to the actions library? I thought this might lead to issues and that's why we are having it in the controller as a mapping?

test/unit-tests/Actions.test.ts
```diff
@@ -182,6 +183,7 @@ contract('Actions', ([owner, random, random2, random3]) => {
 
       const depositArgs = await actionTester.getOpenVaultArgs()
       assert.equal(depositArgs.owner, owner)
+      assert.equal(depositArgs.vaultType, new BigNumber(0))
     })
```
aparnakr, line 187: wanna add a test to make sure type of vault is either 0 or 1?

aparnakr (approved):
Looks good to merge! :D 

(Just something we should test in the future is if upgrading the action lib + controller will have any issues or will it be backward compatible)

contracts/Controller.sol
haythem96: Not sure if it is important, just to make sure it is clear that all those new variables are in order for storage slot

contracts/Controller.sol
```diff
@@ -60,10 +60,17 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
...
+
+    /// @dev mapping to map vault by each vault type, naked margin vault should be set to 1, spread/max loss vault should be set to 0
+    mapping(address => mapping(uint256 => uint256)) internal vaultType;
+    /// @dev mapping to store the timestamp at which the vault was last updated, will be updated in every action that changes the vault state or when calling sync()
```
antoncoding, line 67: where is the `sync` function?

antoncoding (approved):

contracts/Controller.sol
haythem96: In another PR :p

aparnakr (approved):

PR #369, opened 2021-03-08
Update chainlink interface to V3

haythem96:
# Task: Update chainlink interface to V3

## High Level Description

This PR update Chainlink interface from V2 to V3.

### Code

- [x] Unit test 100% coverage

### Documentation

### DIFF
--- a/contracts/interfaces/AggregatorInterface.sol
+++ b/contracts/interfaces/AggregatorInterface.sol
@@ -7,16 +7,34 @@ pragma solidity 0.6.10;
  * @dev Interface of the Chainlink aggregator
  */
 interface AggregatorInterface {
-    function latestAnswer() external view returns (int256);
+    function decimals() external view returns (uint8);
 
-    function latestTimestamp() external view returns (uint256);
+    function description() external view returns (string memory);
 
-    function latestRound() external view returns (uint256);
+    function version() external view returns (uint256);
 
-    function getAnswer(uint256 roundId) external view returns (int256);
+    // getRoundData and latestRoundData should both raise "No data present"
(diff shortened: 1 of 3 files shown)


contracts/interfaces/AggregatorInterface.sol
```diff
@@ -7,16 +7,34 @@ pragma solidity 0.6.10;
...
-    event NewRound(uint256 indexed roundId, address indexed startedBy, uint256 startedAt);
+    function latestRoundData()
+        external
+        view
+        returns (
+            uint80 roundId,
+            int256 answer,
+            uint256 startedAt,
+            uint256 updatedAt,
+            uint80 answeredInRound
```
aparnakr, line 38: what is the answered in round vs the roundId?

aparnakr: Do we need some check around that?

aparnakr (dismissed):

antoncoding (changes requested):
In their v3, [the aggregator also return `decimals`. ](https://docs.chain.link/docs/price-feeds-api-reference#decimals)we should use that instead of assuming things will be 8 decimals

haythem96:
> In their v3, [the aggregator also return `decimals`. ](https://docs.chain.link/docs/price-feeds-api-reference#decimals)we should use that instead of assuming things will be 8 decimals

Oh cool, added that.

antoncoding (approved):

aparnakr (approved):

contracts/interfaces/AggregatorInterface.sol
haythem96: Will look into that

PR #371, opened 2021-03-09
Partial collateralization margin check

haythem96:
# Task: Partial collateralization margin check

## High Level Description

- Added `sync()` function to update latest vault update timestamp
- Updated `getVault()` to return also vault type and vault latest update timestamp
- Updated `_settleVault()` to check if vault is not undercollateralized before settling
- Updated `_withdrawCollateral()` to enable withdrawing collateral from vault that have expired short otoken
- Updated `getExcessCollateral()` to take also vault type as arg
- Added margin check in `_getMarginRequired()` for naked margin vault when not expired yet
- Added `_findUpperBoundValue()` to get option upper bound value
- Added owner functions to set collateral dust, product times to expiry, upper bound values, spot shock value and oracle deviation value
- Fix Certora harness contracts

### Code

- [x] Unit test 100% coverage

### Documentation


contracts/Controller.sol
```diff
@@ -751,15 +761,24 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
...
 
-        OtokenInterface otoken = hasShort
-            ? OtokenInterface(vault.shortOtokens[0])
-            : OtokenInterface(vault.longOtokens[0]);
+            if (hasLong) {
+                OtokenInterface longOtoken = OtokenInterface(vault.longOtokens[0]);
+
+                longOtoken.burnOtoken(address(pool), vault.longAmounts[0]);
+            }
+        }
```
antoncoding, line 781: why do we need this {} block?

contracts/MarginCalculator.sol
```diff
@@ -43,19 +36,227 @@ contract MarginCalculator {
...
+     * @param _timeToExpiry option time to expiry timestamp
+     */
+    function setProductTimeToExpiry(
+        address _underlying,
+        address _strike,
+        address _collateral,
+        bool _isPut,
+        uint256 _timeToExpiry
+    ) external onlyOwner {
+        bytes32 productHash = keccak256(abi.encode(_underlying, _strike, _collateral, _isPut));
```
antoncoding, line 115: we should have a OpynUtil contract which has has this `getProductHash` function that all contracts can inherit, instead of computing it in different contracts.

contracts/MarginCalculator.sol
```diff
@@ -43,19 +36,227 @@ contract MarginCalculator {
...
+        uint256[] storage expiryArray = productTimeToExpiry[productHash];
+
+        require(
+            (expiryArray.length == 0) || (_timeToExpiry > expiryArray[expiryArray.length.sub(1)]),
+            "MarginCalculator: expiry array is not in order"
+        );
```
antoncoding, line 121: so we can't update any timestamp in the the middle of the expiryArray array?

contracts/MarginCalculator.sol
```diff
@@ -43,19 +36,227 @@ contract MarginCalculator {
...
+    }
+
+    /**
+     * @notice set oracle deviation (1e27)
+     * @dev can only be called by owner
+     * @param _deviation deviation value
+     */
+    function setOracleDeviation(uint256 _deviation) external onlyOwner {
+        oracleDeviation = _deviation;
+    }
```
antoncoding, line 183: should oracle deviation be one value for each collateral?

contracts/MarginCalculator.sol
```diff
@@ -183,86 +379,207 @@ contract MarginCalculator {
     function _getMarginRequired(MarginVault.Vault memory _vault, VaultDetails memory _vaultDetails)
         internal
         view
-        returns (FPI.FixedPointInt memory)
+        returns (FPI.FixedPointInt memory, FPI.FixedPointInt memory)
```
antoncoding, line 399: add doc description of the second `FixedPointInt` you're returning!

contracts/Controller.sol
haythem96: stack too deep error 😅

contracts/MarginCalculator.sol
haythem96: Yeah good code refactoring, but still same effect in term of computation.

haythem96: For now only MarginCalculator and Whitelist use product hash, to do the refactor need to modify the Whitelist module and redeploy it again. Will keep it as it is for now.

contracts/MarginCalculator.sol
haythem96: Yes

contracts/Controller.sol
haythem96: That is basically a diff scope to declare variables in, in this case the `hasShort` and `hasLong` will not exist outside of that block.

contracts/MarginCalculator.sol
antoncoding: I think we can start using it for the Calculator, it doesn't look good to write this: 
`keccak256(abi.encode(_underlying, _strike, _collateral, _isPut))` every time.

contracts/MarginCalculator.sol
```diff
@@ -43,19 +36,227 @@ contract MarginCalculator {
...
+     * @param _collateral otoken collateral asset
+     * @param _isPut otoken type
+     * @return _shockValue spot shock value (1e27)
+     */
+    function getSpotShock(
+        address _underlying,
+        address _strike,
+        address _collateral,
+        bool _isPut
+    ) external view returns (uint256) {
```
antoncoding, line 246: Does it need to external?

haythem96: Yeah, it is a getter function know what is the spot shock value for a product is, as `spotShock` mapping is internal.

contracts/MarginCalculator.sol
haythem96: It is set as one value for all in the spec.

test/integration-tests/longCallSpreadExpireItm.test.ts
```diff
@@ -271,7 +271,7 @@ contract('Long Call Spread Option expires Itm flow', ([accountOwner1, nakedBuyer
 
       // Check that we start at a valid state
       const vaultBefore = await controllerProxy.getVault(accountOwner1, vaultCounter1)
-      const vaultStateBefore = await calculator.getExcessCollateral(vaultBefore)
+      const vaultStateBefore = await calculator.getExcessCollateral(vaultBefore[0], vaultBefore[1])
```
antoncoding, line 274: Can u re-write this as
```javascript
const [vaultBefore, vaultType] = await controllerProxy.getVault(accountOwner1, vaultCounter1)
const vaultStateBefore = await calculator.getExcessCollateral(vaultStateBefore, vaultType)
```

or 

```javascript
const [vaultBefore, vaultType] = await controllerProxy.getVault(accountOwner1, vaultCounter1)
const [amount, isExcess] = await calculator.getExcessCollateral(vaultStateBefore, vaultType)
```

Using [0] and [1] here on both result of `getVault` (which represent collateral & type) and `getExcessCollateral` (which represent `amount`, `isExcess`) is very confusing

test/integration-tests/longCallSpreadExpireItm.test.ts
```diff
@@ -339,22 +339,22 @@ contract('Long Call Spread Option expires Itm flow', ([accountOwner1, nakedBuyer
 
       // Check that we end at a valid state
       const vaultAfter = await controllerProxy.getVault(accountOwner1, vaultCounter1)
-      const vaultStateAfter = await calculator.getExcessCollateral(vaultAfter)
+      const vaultStateAfter = await calculator.getExcessCollateral(vaultAfter[0], vaultAfter[1])
       assert.equal(vaultStateAfter[0].toString(), '0')
       assert.equal(vaultStateAfter[1], true)
```
antoncoding, line 344: similarly, I think it's better to use 
```javascript
const [vaultAfter, ] = await controllerProxy.getVault(accountOwner1, vaultCounter1)
```

so u don't have to use `vaultAfter[0]`

contracts/Controller.sol
antoncoding: wow this is a cool hack!

test/integration-tests/longCallSpreadExpireItm.test.ts
haythem96: This need a refactor in different integration files, will do it later in another PR!

contracts/MarginCalculator.sol
haythem96: Added.

contracts/libs/Actions.sol
```diff
@@ -152,8 +152,13 @@ library Actions {
         require(_args.actionType == ActionType.OpenVault, "Actions: can only parse arguments for open vault actions");
         require(_args.owner != address(0), "Actions: cannot open vault for an invalid account");
 
-        // decode vault type from _args.data
-        uint256 vaultType = abi.decode(_args.data, (uint256));
+        // if not _args.data included, vault type will be 0 by default
```
haythem96, line 155: This should not break project that already built on top of Gamma, when using the openVault action

test/integration-tests/longCallSpreadExpireItm.test.ts
antoncoding: can u make an issue for it and make sure it happen? don't want to increase our technical debt 😸

haythem96: mmm not sure it is a technical debt even, as it is just another way or writing! but yeah will open an issue.

antoncoding (approved):

contracts/Controller.sol
haythem96: Will un-resolve this for now so that other can see this...

test/unit-tests/controller.test.ts
```diff
@@ -2701,29 +2699,30 @@ contract(
...
+          //       index: '0',
+          //       data: ZERO_ADDR,
+          //     },
+          //   ]
+
+          //   await expectRevert(
+          //     controllerProxy.operate(actionArgs, {from: accountOwner1}),
+          //     'Controller: can not withdraw collateral from a vault with an expired short otoken',
+          //   )
+          // })
```
aparnakr, line 2725: Can you test that if you withdraw collateral after expiry, you cannot withdraw more than the amount that settleVault would give you? (For a type 0 vault)

contracts/Controller.sol
```diff
@@ -650,16 +670,6 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
...
-        MarginVault.Vault memory vault = getVault(_args.owner, _args.vaultId);
-        if (_isNotEmpty(vault.shortOtokens)) {
-            OtokenInterface otoken = OtokenInterface(vault.shortOtokens[0]);
-
-            require(
-                now < otoken.expiryTimestamp(),
-                "Controller: can not withdraw collateral from a vault with an expired short otoken"
-            );
-        }
-
```
aparnakr, line 662: Why do we need to remove this and open it up after expiry? :) Just trying to think about what case needs this!

haythem96: I removed this so vault owner can withdraw remaining collateral from vault after liquidation. And added checking margin vault after withdraw collateral action.

contracts/Controller.sol
aparnakr: This is cool! :D

contracts/Controller.sol
haythem96: Added tests for withdrawing more collateral than payout in an expired position, withdrawing some of the collateral and settling should still only give a sum of amounts equal to the payout https://github.com/opynfinance/GammaProtocol/blob/6647517cf4bd30de0b3bf763c70e620a450e223b/test/unit-tests/controller.test.ts#L3640

test/unit-tests/controller.test.ts
haythem96: Added tests for withdrawing more collateral than payout in an expired position, withdrawing some of the collateral and settling should still only give a sum of amounts equal to the payout

https://github.com/opynfinance/GammaProtocol/blob/6647517cf4bd30de0b3bf763c70e620a450e223b/test/unit-tests/controller.test.ts#L3640

contracts/Controller.sol
aparnakr: if a vault got liquidated they can still withdraw through add collateral right?

aparnakr: would prefer if we kept this check, just want to think through the issues this could cause before opening this up

test/unit-tests/marginCalculatorNakedMargin.test.ts
```diff
@@ -0,0 +1,848 @@
...
+  // if (isPut) {
+  //   a = BigNumber.minimum(strikePrice, spotShockValue.times(underlyingPrice).dividedBy(1e27))
+  //   b = BigNumber.maximum(strikePrice.minus(spotShockValue.times(underlyingPrice).dividedBy(1e27)), new BigNumber(0))
+  //   marginRequired = upperBoundValue.times(a).plus(b).times(shortAmount).dividedBy(1e27)
+  // } else {
+  //   const one = scaleBigNum(1, 27)
+  //   a = BigNumber.minimum(one, strikePrice.dividedBy(underlyingPrice.dividedBy(spotShockValue)))
+  //   b = BigNumber.maximum(one.minus(strikePrice.dividedBy(underlyingPrice.dividedBy(spotShockValue))), new BigNumber(0))
+  //   marginRequired = upperBoundValue.times(a).plus(b).times(shortAmount);
+  // }
```
aparnakr, line 44: why is this commented out?

haythem96: need to remove it, not using that code, using the one above.

test/unit-tests/marginCalculatorNakedMargin.test.ts
```diff
@@ -0,0 +1,848 @@
...
+  let dai: MockERC20Instance
+  let weth: MockERC20Instance
+
+  const usdcDecimals = 6
+  const daiDecimals = 8
+  const wethDecimals = 18
+
+  const vaultType = 1
+
+  const errorDelta = 0.1
```
aparnakr, line 87: where is the error delta coming from? Is it js or solidity?

haythem96: there is a rounding issue in the get margin function for the naked vault case, so I explicitly rounded up for the getMargin, and making sure the rounding issue is not bigger than 0.1 for a position of 1 size. (it is a solidity issue)

haythem96: Planning to explicitly round up for returning the margin required and round down for the liquidation price.

test/unit-tests/marginCalculatorNakedMargin.test.ts
```diff
@@ -0,0 +1,848 @@
...
+      const isPut = true
+      const optionExpiry = new BigNumber(await time.latest()).plus(timeToExpiry[0])
+      // get option upper bound value
+      const upperBoundValue = new BigNumber(
+        await calculator.findUpperBoundValue(weth.address, usdc.address, usdc.address, true, optionExpiry),
+      )
+        .dividedBy(1e27)
+        .toNumber()
+
+      // console.log('Upper boudn value found: ', upperBoundValue.toString())
```
aparnakr, line 743: Can you remove all console.logs? :)

test/unit-tests/marginCalculatorNakedMargin.test.ts
```diff
@@ -0,0 +1,848 @@
...
+      )
+
+      // assert.isAtMost(
+      //   calcRelativeDiff(new BigNumber('16777.78677'), requiredMargin.dividedBy(10 ** wethDecimals)).toNumber(),
+      //   errorDelta,
+      //   'big error delta',
+      // )
+    })
+
+    it('should return required margin for naked margin vault: 100$ WETH put option with 150 spot price and 2 weeks to expiry', async () => {
```
aparnakr, line 555: could you also test it calculates it right for ITM calls and ITM puts? :D

aparnakr (approved):

test/unit-tests/marginCalculatorNakedMargin.test.ts
haythem96: yup will remove them later,

test/unit-tests/marginCalculatorNakedMargin.test.ts
haythem96: Ah u meant ITM instead of OTM, I guess it is the same calculation, IMO this cover the basics and will add more in integration tests.

aparnakr (approved):

contracts/Controller.sol
```diff
@@ -389,9 +401,9 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
      * @return amount of collateral that can be taken out
      */
     function getProceed(address _owner, uint256 _vaultId) external view returns (uint256) {
-        MarginVault.Vault memory vault = getVault(_owner, _vaultId);
+        (MarginVault.Vault memory vault, uint256 typeVault, ) = getVault(_owner, _vaultId);
```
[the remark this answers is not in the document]
haythem96: Added vault type to specify if vault is a naked margin vault, or max loss/spread vault (`1` or `0`)

contracts/Controller.sol
```diff
@@ -769,13 +788,9 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
             "Controller: asset prices not finalized yet"
         );
 
-        (uint256 payout, ) = calculator.getExcessCollateral(vault);
+        (uint256 payout, bool isValidVault) = calculator.getExcessCollateral(vault, typeVault);
```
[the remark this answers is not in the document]
haythem96: Didn't change anything here, just re-ordered instructions and used a new `{}` scope to avoid stack too deep error.

contracts/libs/Actions.sol
```diff
@@ -152,8 +152,13 @@ library Actions {
...
+        uint256 vaultType;
+
+        if (_args.data.length == 32) {
```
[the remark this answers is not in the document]
haythem96: Yeah need to add more commenting, but basically here after this upgrade, user need to specify the vault type when opening vault (0 for max loss/spread and 1 for naked margin), so the vault type will best sent in the _args.data as an encoded uint256 (32bits).

specs/harness/ControllerHarness.sol
```diff
@@ -81,7 +81,8 @@ contract ControllerHarness is Controller {
     (, bool isValidVault) = calculator.getExcessCollateral(
       _vault.shortAmounts[0],
       _vault.longAmounts[0],
-      _vault.collateralAmounts[0]
+      _vault.collateralAmounts[0],
```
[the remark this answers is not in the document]
haythem96: IMO a more simplified contract for Certora rules... cc @aparnakr

contracts/Controller.sol
haythem96: Yeah they can still do that if the short otoken is not expired yet through _withdrawCollateral action, if the short is expired, they settle vault and will get the proceed and the excess amount also.

PR #373, opened 2021-03-16
fix the controller harness

aparnakr:
# Task: Fix no bankruptcy failing tests
## High Level Description

Specific Changes
Changed the controller harness openVault call to accept 3 args

### Code

- [x] Unit test 100% coverage

### Documentation

### DIFF
--- a/specs/harness/ControllerHarness.sol
+++ b/specs/harness/ControllerHarness.sol
@@ -150,7 +150,7 @@ contract ControllerHarness is Controller {
   }
 
   function openVault(address owner, uint256 vaultId) external {
-    Actions.OpenVaultArgs memory args = Actions.OpenVaultArgs({owner: owner, vaultId: vaultId});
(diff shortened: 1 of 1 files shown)


specs/harness/ControllerHarness.sol
```diff
@@ -150,7 +150,7 @@ contract ControllerHarness is Controller {
   }
 
   function openVault(address owner, uint256 vaultId) external {
-    Actions.OpenVaultArgs memory args = Actions.OpenVaultArgs({owner: owner, vaultId: vaultId});
+    Actions.OpenVaultArgs memory args = Actions.OpenVaultArgs({owner: owner, vaultId: vaultId, vaultType: 0});
```
haythem96, line 153: Shouldn't this be dynamic somehow ?

haythem96 (approved):

specs/harness/ControllerHarness.sol
aparnakr: well for now just having this test on vaultType 0. For margin vaults, might need to write a new test. When I do that, will update this!

PR #378, opened 2021-03-24
Opyn liquidation

haythem96:
# Task: Opyn liquidation 🔥

## High Level Description

This PR add liquidation feature for the partial collateralization upgrade.

**Controller.sol**
- Add `isLiquidatable()` to check if a specific vault is liquidatable at a specific chainlink round timestamp
- Add `_liquidate()` function and include in the operate call.

**Calculator.sol**
- Add `isLiquidatable()` to check vault liquidation
- Add `_price()` to get liquidation price
- Rename `getVaultDetails()` to `_getVaultDetails()`

**Actions.sol**
- Add `LiquidateArgs` and `_parseLiquidateArgs()`

### Code

- [x] Unit test 100% coverage

### Documentation


contracts/Controller.sol
```diff
@@ -780,16 +857,22 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
             }
         }
 
+        // get otoken details
```
antoncoding, line 860: I don't think we need comments on every line of the code. The code itself should be self-explaining, comments like this `get otoken details` while the function name is called `getOtokenDetails` is just redundant IMO.

contracts/Controller.sol
```diff
@@ -799,6 +882,62 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
...
+        // the price is the amount of collateral asset to pay per 1 repaid debt(otoken)
+        // collateralDust is the amount of collateral that should be left in vault if not fully liquidated by liquidator
+        (MarginVault.Vault memory vault, bool isLiquidatable, uint256 price, uint256 collateralDust) = _isLiquidatable(
+            _args.owner,
+            _args.vaultId,
+            _args.roundId
+        );
+
+        // do not allow liquidating un-liquidatable vault
+        require(isLiquidatable, "Controller: can not liquidate vault");
```
antoncoding, line 903:
```suggestion
        require(isLiquidatable, "Controller: can not liquidate the vault");
```

antoncoding: // do not allow liquidating un-liquidatable vault
this comment is redundant too

contracts/Controller.sol
```diff
@@ -799,6 +882,62 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
...
+
+        // amount of collateral to offer to liquidator
+        uint256 collateralToSell = _args.amount.mul(price).div(1e8);
+
+        // burn short otoken from liquidator address, index of short otoken hardcoded at 0
+        // should always work, if vault have no short otoken, it will not reach this step
```
antoncoding, line 909:
```suggestion
        // if a vault have no short otoken, it will not reach this step
```

contracts/Controller.sol
```diff
@@ -799,6 +882,62 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
...
+        pool.transferToUser(vault.collateralAssets[0], _args.receiver, collateralToSell);
+
+        // if vault is partially liquidated, amount of short otoken still greater than zero
+        // make sure remaining collateral amount is greater than dust amount
+        if (vault.shortAmounts[0].sub(_args.amount) > 0) {
+            require(
+                vault.collateralAmounts[0].sub(collateralToSell) >= collateralDust,
+                "Controller: can not leave less than collateral dust"
+            );
+        }
```
antoncoding, line 928: we should do the check earlier in the function to meet our pattern.

contracts/Controller.sol
```diff
@@ -838,6 +977,38 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
         return whitelist.isWhitelistedCallee(_callee);
     }
 
+    /**
+     * @notice check if a vault is liquidatable in a specific round id
+     * @param _owner vault owner address
+     * @param _vaultId vault id to check
+     * @param _roundId chainlink round id to check vault status at
+     * @return vault struct, isLiquidatable, true if vault is undercollateralized, the price of 1 repaid otoken and the otoken collateral dust amount
```
antoncoding, line 985: also explain that price is scaled at 1e18. 
Also why 1e18?

test/utils.ts
```diff
@@ -154,3 +156,58 @@ export const signOrder = async (signer: any, order: any) => {
...
+
+export const calcRelativeDiff = (expected: BigNumber, actual: BigNumber): BigNumber => {
+  let diff: BigNumber
+
+  if (actual.isGreaterThan(expected)) {
+    diff = actual.minus(expected)
+  } else {
+    diff = expected.minus(actual)
+  }
+  return diff
```
antoncoding, line 212: I think there's an `.abs()` function, so you can use `actual.minus(expected).abs()`

contracts/MarginCalculator.sol
```diff
@@ -112,18 +119,24 @@ contract MarginCalculator is Ownable {
...
+        // get array of expiries
         uint256[] storage expiryArray = productTimeToExpiry[productHash];
 
+        // check that this is the first expiry to set, if not the last expiry should be less than the new one to insert (to make sure the array stay in order)
         require(
             (expiryArray.length == 0) || (_timeToExpiry > expiryArray[expiryArray.length.sub(1)]),
             "MarginCalculator: expiry array is not in order"
         );
+
+        // check that this time to expiry have an upper bound value (setTimeToExpiryValue() should be called before this)
```
antoncoding, line 133: can we ask Andrew to rename this variable? `timeToExpiryValue` is causing confusion as it is so meaningless

contracts/MarginCalculator.sol
```diff
@@ -168,8 +184,10 @@ contract MarginCalculator is Ownable {
         bool _isPut,
         uint256 _shockValue
     ) external onlyOwner {
+        // get product hash
```
antoncoding, line 187:
```suggestion
```

contracts/MarginCalculator.sol
```diff
@@ -168,8 +184,10 @@ contract MarginCalculator is Ownable {
...
         bytes32 productHash = _getProductHash(_underlying, _strike, _collateral, _isPut);
 
+        // set spot shock value
         spotShock[productHash] = _shockValue;
```
antoncoding, line 191: should describe what scale this value is in the function comment.

contracts/MarginCalculator.sol
```diff
@@ -281,12 +299,17 @@ contract MarginCalculator is Ownable {
         uint256 _collateralDecimals,
         bool _isPut
     ) external view returns (uint256) {
+        // get product hash
         bytes32 productHash = _getProductHash(_underlying, _strike, _collateral, _isPut);
 
+        // scale short amount from 1e8 to 1e27
         FPI.FixedPointInt memory shortAmount = FPI.fromScaledUint(_shortAmount, BASE);
+        // scale short strike from 1e8 to 1e27
         FPI.FixedPointInt memory shortStrike = FPI.fromScaledUint(_strikePrice, BASE);
```
antoncoding, line 308: These comments are not helpful. Either rename the `fromScaledUint` function name to make it self explaining or remove the comments.

contracts/MarginCalculator.sol
```diff
@@ -345,6 +368,128 @@ contract MarginCalculator is Ownable {
...
+    // struct to store isLiquidatable and the price for 1 otoken repaid scaled to the collateral decimals
+    struct LiquidationStatus {
+        bool isLiquidatable;
+        uint256 debtPrice;
+    }
+
+    /**
+     * @notice check if a specific vault is undercollateralized at a specific chainlink round
+     */
+    function isLiquidatable(
```
antoncoding, line 387: add function params and return description

contracts/MarginCalculator.sol
```diff
@@ -345,6 +368,128 @@ contract MarginCalculator is Ownable {
...
+        external
+        view
+        returns (
+            bool,
+            uint256,
+            uint256
+        )
+    {
+        // create LiquidationStatus struct to avoid stack too deep error
+        // will store isLiquidatable and the liquidation price
```
antoncoding, line 402:
```suggestion
```
We don't need to explain again what this struct is storing.

contracts/MarginCalculator.sol
```diff
@@ -345,6 +368,128 @@ contract MarginCalculator is Ownable {
...
+        LiquidationStatus memory liquidationStatus = LiquidationStatus(false, 0);
+
+        // liquidation is only supported for naked margin vault
+        require(_vaultType == 1, "MarginCalculator: invalid vault type to liquidate");
+
+        // get vault details
```
antoncoding, line 408:
```suggestion
```

contracts/MarginCalculator.sol
```diff
@@ -345,6 +368,128 @@ contract MarginCalculator is Ownable {
...
+        VaultDetails memory vaultDetails = _getVaultDetails(_vault, _vaultType);
+
+        // can not liquidate vault that have no short position
+        if (!vaultDetails.hasShort) return (false, 0, 0);
+
+        // do not allow liquidating expired position
+        require(
+            now < vaultDetails.shortExpiryTimestamp,
+            "MarginCalculator: can not liquidate vault with expired short otoken"
+        );
```
antoncoding, line 418: when are we returning (false, 0, 0) and when are we reverting?

contracts/MarginCalculator.sol
```diff
@@ -345,6 +368,128 @@ contract MarginCalculator is Ownable {
...
+        (uint256 price, uint256 timestamp) = oracle.getChainlinkRoundData(
+            vaultDetails.shortUnderlyingAsset,
+            uint80(_roundId)
+        );
+
+        // check that price timestamp is after latest timestamp the vault was updated at
+        require(
+            timestamp >= _vaultLatestUpdate,
+            "MarginCalculator: auction timestamp should be post vault latest update"
+        );
```
antoncoding, line 430: again, why not just return `false`? 
I think it's better to avoid using revert in a view functions, so it's easier to other contracts to integrate with. 
It's kind of inconsistent calling a function will sometime cause reverts and sometime return false.

contracts/MarginCalculator.sol
```diff
@@ -345,6 +368,128 @@ contract MarginCalculator is Ownable {
...
+
+        // another struct to store some useful short otoken details, to avoid stack to deep error
+        ShortScaledDetails memory shortDetails = ShortScaledDetails({
+            shortAmount: FPI.fromScaledUint(_vault.shortAmounts[0], BASE),
+            shortStrike: FPI.fromScaledUint(vaultDetails.shortStrikePrice, BASE),
+            shortUnderlyingPrice: FPI.fromScaledUint(price, BASE)
+        });
+
+        // get product hash
```
antoncoding, line 439:
```suggestion
```

contracts/MarginCalculator.sol
```diff
@@ -345,6 +368,128 @@ contract MarginCalculator is Ownable {
...
+
+        // convert vault collateral to a fixed point (1e27) from collateral decimals
+        FPI.FixedPointInt memory collateralAmount = FPI.fromScaledUint(
+            _vault.collateralAmounts[0],
+            vaultDetails.collateralDecimals
+        );
+
+        // another scope to avoid stack too deep!
+        {
+            // get required collateral
```
antoncoding, line 455:
```suggestion
```

contracts/MarginCalculator.sol
```diff
@@ -345,6 +368,128 @@ contract MarginCalculator is Ownable {
...
+                productHash,
+                shortDetails.shortAmount,
+                shortDetails.shortStrike,
+                shortDetails.shortUnderlyingPrice,
+                vaultDetails.shortExpiryTimestamp,
+                vaultDetails.isShortPut
+            );
+
+            // check if deposited collateral is greater or equal than needed collateral
+            liquidationStatus.isLiquidatable = collateralRequired.isGreaterThan(collateralAmount);
```
antoncoding, line 466: better rename the `collateralAmount` variable to sometime more self explaining, like `depositedCollateral` so you don't need to explicitly say what you're comparing.

contracts/MarginCalculator.sol
```diff
@@ -756,7 +916,87 @@ contract MarginCalculator is Ownable {
...
+        FPI.FixedPointInt memory _vaultCollateral,
+        FPI.FixedPointInt memory _vaultDebt,
+        FPI.FixedPointInt memory _cv,
+        FPI.FixedPointInt memory _spotPrice,
+        uint256 _auctionStartingTime,
+        uint256 _collateralDecimals,
+        bool _isPut
+    ) internal view returns (uint256) {
+        // price of 1 repaid otoken in collateral asset, scaled to collateral decimals and explicitly rounded down
+        FPI.FixedPointInt memory price;
```
antoncoding, line 956: I think it's better to move this declaration right before variable assignment. (right before the if block where you calculate price). otherwise it's weird to see all the definition but just seeing
```
FPI.FixedPointInt memory price;
```

contracts/MarginCalculator.sol
```diff
@@ -756,7 +916,87 @@ contract MarginCalculator is Ownable {
...
+            // store auctionElapsedTime in a FixedPointInt scaled by 1e27
+            FPI.FixedPointInt memory auctionElapsedTime = FPI.fromScaledUint(auctionElapsedTime, 18);
+            // store AUCTION_TIME in a FixedPointInt (already scaled by 1e27)
+            FPI.FixedPointInt memory auctionTime = FPI.fromScaledUint(AUCTION_TIME, 18);
+
+            // calculate price of 1 repaid otoken, scaled by the collateral decimals, expilictly rounded down
+            price = startingPrice.add((endingPrice.sub(startingPrice)).mul(auctionElapsedTime).div(auctionTime));
+        }
+
+        return price.toScaledUint(_collateralDecimals, true);
```
antoncoding, line 990: I think we will need to scales this price variable by another 1e8 or 1e18, instead of just collateral decimal? Otherwise if the collateral asset has few decimals (for example of collateral is eth but have 0 decimals), 1 otoken will always worth 0. (price will always be 0).

contracts/MarginCalculator.sol
```diff
@@ -756,7 +916,87 @@ contract MarginCalculator is Ownable {
...
+     * @param _spotPrice option underlying asset price (in USDC)
+     * @param _auctionStartingTime auction starting timestamp (_spotPrice timestamp from chainlink)
+     * @param _collateralDecimals collateral asset decimals
+     * @param _isPut otoken type, true for put, false for call option
+     * @return price of 1 debt otoken in collateral asset scaled by collateral decimals
+     */
+    function _price(
+        FPI.FixedPointInt memory _vaultCollateral,
+        FPI.FixedPointInt memory _vaultDebt,
+        FPI.FixedPointInt memory _cv,
```
antoncoding, line 949: Why using abbreviation only here 😂 . I think it's better to make it consistent as we are using `getCashValue` elsewhere

antoncoding (changes requested):
Haven't looked closely into the tests yet, will do that later when the CICD passed. 
Generally I think the comments are a bit too much 😂 , I'm leaning toward remove most of them and try to make our code self-explaining.

contracts/Controller.sol
antoncoding: I assume it's 1e18 because you hardcoded a .div(1e18) in the collateral amount calculation

contracts/MarginCalculator.sol
haythem96: Can u explain more ? "(for example of collateral is eth but have 0 decimals)" ?

contracts/MarginCalculator.sol
haythem96: why do u think it is not helpful ?

contracts/Controller.sol
haythem96: what is scaled by 1e8 ? the short amount is in 8decimals, that's why I divided by 1e8 in the end.

haythem96: The liquidation price is always in collateral short otoken collateral decimals.

contracts/MarginCalculator.sol
haythem96: Open a thread about that, can change it later in another PR.

contracts/MarginCalculator.sol
haythem96: will rename it to `availableCollateral`

contracts/Controller.sol
aparnakr: unless the comments explain more than what the code says ;)

contracts/libs/Actions.sol
```diff
@@ -281,6 +295,25 @@ library Actions {
         return SettleVaultArgs({owner: _args.owner, vaultId: _args.vaultId, to: _args.secondAddress});
     }
 
+    function _parseLiquidateArgs(ActionArgs memory _args) internal pure returns (LiquidateArgs memory) {
+        require(_args.actionType == ActionType.Liquidate, "Actions: can only parse arguments for liquidate action");
+        require(_args.owner != address(0), "Actions: cannot liquidate vault for an invalid account owner");
+        require(_args.secondAddress != address(0), "Actions: cannot send collateral to an invalid account");
+        require(_args.data.length == 32, "Actions: cannot parse liquidate action with no round id");
+
```
aparnakr, line 303: Do we also want a check to not allow self liquidation?

contracts/Controller.sol
```diff
@@ -799,6 +880,61 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
...
+        // make sure remaining collateral amount is greater than dust amount
+        if (vault.shortAmounts[0].sub(_args.amount) > 0) {
+            require(
+                vault.collateralAmounts[0].sub(collateralToSell) >= collateralDust,
+                "Controller: can not leave less than collateral dust"
+            );
+        }
+
+        // burn short otoken from liquidator address, index of short otoken hardcoded at 0
+        // should always work, if vault have no short otoken, it will not reach this step
```
aparnakr, line 915: love all these comments! :D Makes our code much more readable!!

contracts/libs/Actions.sol
haythem96: We currently allow for self liquidation

haythem96: thoughts @aleone @zkoticha ?

contracts/MarginCalculator.sol
```diff
@@ -345,6 +368,129 @@ contract MarginCalculator is Ownable {
...
+        FPI.FixedPointInt shortAmount;
+        FPI.FixedPointInt shortStrike;
+        FPI.FixedPointInt shortUnderlyingPrice;
+    }
+    // struct to store isLiquidatable and the price for 1 otoken repaid scaled to the collateral decimals
+    struct LiquidationStatus {
+        bool isLiquidatable;
+        uint256 debtPrice;
+    }
+
```
aparnakr, line 385: this is smart :D

contracts/MarginCalculator.sol
```diff
@@ -345,6 +368,129 @@ contract MarginCalculator is Ownable {
...
+    /**
+     * @notice check if a specific vault is undercollateralized at a specific chainlink round
+     * @param _vault vault struct
```
aparnakr, line 388:
```suggestion
     * @dev if the vault is of type 0, the function will revert. 
```

test/unit-tests/liquidations.test.ts
```diff
@@ -0,0 +1,442 @@
...
+        weth.address,
+        usdc.address,
+        usdc.address,
+        scaledStrikePrice,
+        optionExpiry,
+        isPut,
+      )
+    })
+
+    it('should not be able to liquidate vault with type equal to 1', async () => {
```
aparnakr, line 143: it should be able to liquidate this right?

test/utils.ts
```diff
@@ -154,3 +154,53 @@ export const signOrder = async (signer: any, order: any) => {
   order.signature = `0x${v}${rs}02`
   return order
 }
+
+BigNumber.config({ROUNDING_MODE: BigNumber.ROUND_DOWN})
+
+export const expectedLiqudidationPrice = (
+  collateral: number | string,
+  debt: number,
```
aparnakr, line 162: this is good to use this instead of hardcoding values! :D

test/unit-tests/liquidations.test.ts
haythem96: Yup a typo mistake 😅

aparnakr (approved):

antoncoding (approved):

PR #382, opened 2021-04-12
Partial Collateralization & Liquidation

haythem96:
# Gamma Partial Collateralization

### Code

- [x] Unit test 100% coverage

### Documentation


contracts/Controller.sol
```diff
@@ -339,7 +356,22 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
...
+    }
+
+    /**
+     * @notice sync vault latest update timestamp
+     * @dev anyone can update the latest time the vault was touched by calling this function
+     * vaultLatestUpdate will sync if the vault is well collateralized
+     * @param _owner vault owner address
+     * @param _vaultId vault id
+     */
+    function sync(address _owner, uint256 _vaultId) external nonReentrant notFullyPaused {
```
aleone, line 372: Is it useful to have a canSync public function or just do via f/e calls or require devs monitoring via bots to do it themselves?

contracts/MarginCalculator.sol
```diff
@@ -16,19 +17,20 @@ import {MarginVault} from "./libs/MarginVault.sol";
...
+    /// @dev decimals option upper bound value, spot shock and oracle deviation
+    uint256 internal constant SCALING_FACTOR = 27;
 
     /// @dev decimals used by strike price and oracle price
     uint256 internal constant BASE = 8;
 
-    /// @dev FixedPoint 0
-    FPI.FixedPointInt internal ZERO = FPI.fromScaledUint(0, BASE);
+    /// @notice auction length
+    uint256 public constant AUCTION_TIME = 3600;
```
aleone, line 31: I lean towards this being adjustable.  If we start having failed liquidations we may need to adjust this quickly.

I guess we could always deploy a new controller, set all the values and then migrate it in the address book.  That would require a bunch of tx, but only one tx to migrate, so we don't have a broken/controller with non finalized values.

aleone: If we do allow it to be set, we could set a min value - admin can prb steal a lot of margined funds by atomically:

-flashloan bunch of eth
-changing margin requirements
-setting auction_time to 0
-liquidating a bunch of folks (they might need a recursive operate function to do this have to mint+liquidate+burn)
-setting margin req back to normal + auction time back to normal

contracts/MarginCalculator.sol
```diff
@@ -43,19 +45,288 @@ contract MarginCalculator {
...
+        // set spot shock value in 27 decimals
+        spotShock[productHash] = _shockValue;
+    }
+
+    /**
+     * @notice set oracle deviation (1e27)
+     * @dev can only be called by owner
+     * @param _deviation deviation value
+     */
+    function setOracleDeviation(uint256 _deviation) external onlyOwner {
```
aleone, line 201: Might make sense this should be a mapping per product as chainlink deviations are different between products.  COMPUSD is 1% ETHUSD 0.5%, LINKUSD 1%, etc.

We could also just be conservative and use 2% or something.

contracts/MarginCalculator.sol
```diff
@@ -43,19 +45,288 @@ contract MarginCalculator {
...
+    uint256 internal oracleDeviation;
+
+    /// @dev FixedPoint 0
+    FPI.FixedPointInt internal ZERO = FPI.fromScaledUint(0, BASE);
+
+    /// @dev mapping to store dust amount per option collateral asset (scaled by collateral decimals)
+    mapping(address => uint256) internal dust;
+
+    /// @dev mapping to store array of time to expiry per product
+    mapping(bytes32 => uint256[]) internal productTimeToExpiry;
```
aleone, line 66: Think we should brainstorm variable names for this and timeToExpiryValue- think is a bit confusing.

contracts/MarginCalculator.sol
```diff
@@ -43,19 +45,288 @@ contract MarginCalculator {
...
+    function getTimeToExpiryValue(
+        address _underlying,
+        address _strike,
+        address _collateral,
+        bool _isPut,
+        uint256 _timeToExpiry
+    ) external view returns (uint256) {
+        bytes32 productHash = _getProductHash(_underlying, _strike, _collateral, _isPut);
+
+        return timeToExpiryValue[productHash][_timeToExpiry];
```
aleone, line 250: maybe we should have a function that retuns all the values and all of the timeToExpiry via looping through productTimeToExpiry

This is similar to the comment about adding things 1 by 1.  Might be useful to return all for a given product as well as add all for a given product.

contracts/MarginCalculator.sol
```diff
@@ -99,17 +370,146 @@ contract MarginCalculator {
...
+            bool,
+            uint256,
+            uint256
+        )
+    {
+        // create LiquidationStatus struct to avoid stack too deep error
+        LiquidationStatus memory liquidationStatus = LiquidationStatus(false, 0);
+
+        // liquidation is only supported for naked margin vault
+        require(_vaultType == 1, "MarginCalculator: invalid vault type to liquidate");
```
aleone, line 413: Don't have strong opinions about reverting this view function - but should discuss how we will program a liquidation bot to provide to community and run ourselves and if reverts on view functions will be unexpected or cause issues.

I understand the rationale for providing something other than false when a vault can never be liquidated vs just not liquidatable now.

contracts/Controller.sol
```diff
@@ -773,6 +880,61 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
...
+     * @notice liquidate naked margin vault
+     * @dev can liquidate different vaults id in the same operate() call
+     * @param _args liquidation action arguments struct
+     */
+    function _liquidate(Actions.LiquidateArgs memory _args) internal notPartiallyPaused {
+        require(_checkVaultId(_args.owner, _args.vaultId), "Controller: invalid vault id");
+
+        // check if vault is undercollateralized
+        // the price is the amount of collateral asset to pay per 1 repaid debt(otoken)
+        // collateralDust is the amount of collateral that should be left in vault if not fully liquidated by liquidator
```
aleone, line 893: collateralDust is the minimum amount of collateral that can be left in the vault when a partial liquidation occurs

contracts/libs/Actions.sol
```diff
@@ -150,7 +166,18 @@ library Actions {
         require(_args.actionType == ActionType.OpenVault, "Actions: can only parse arguments for open vault actions");
         require(_args.owner != address(0), "Actions: cannot open vault for an invalid account");
 
-        return OpenVaultArgs({owner: _args.owner, vaultId: _args.vaultId});
+        // if not _args.data included, vault type will be 0 by default
+        uint256 vaultType;
```
aleone, line 170: Like this for backward compatability!

aleone, line 170: Like this for non breaking changes!

contracts/pricers/ChainlinkPricer.sol
```diff
@@ -53,30 +61,56 @@ contract ChainLinkPricer is OpynPricerInterface {
...
     /**
      * @notice get the live price for the asset
      * @dev overides the getPrice function in OpynPricerInterface
      * @return price of the asset in USD, scaled by 1e8
      */
     function getPrice() external override view returns (uint256) {
-        int256 answer = aggregator.latestAnswer();
+        (, int256 answer, , , ) = aggregator.latestRoundData();
         require(answer > 0, "ChainLinkPricer: price is lower than 0");
         // chainlink's answer is already 1e8
```
aleone, line 86: prb delete this comment

contracts/MarginCalculator.sol
```diff
@@ -528,10 +1067,163 @@ contract MarginCalculator {
...
+        bool _isPut
+    ) internal pure returns (bytes32) {
+        return keccak256(abi.encode(_underlying, _strike, _collateral, _isPut));
+    }
+
+    /**
+     * @notice get option cash value
+     * @dev this assume that the underlying price is denominated in strike asset
+     * for put, cv = strike price - underlying price
+     * for call, cv = underlting price - strike price
```
aleone, line 1215: underlying
prb should use below for clarity:
max(underlying price - strike price, 0)

same for puts

contracts/MarginCalculator.sol
```diff
@@ -528,10 +1067,163 @@ contract MarginCalculator {
...
+     * @param _strikePrice option srike price
```
aleone, line 1216: strike

contracts/Controller.sol
haythem96: what `canSync()` for ? to check if it is able to sync? if so it will be basically the same as `sync()` that will check the margin and return bool. We can keep it in the frontend as it can be done by checking the margin state.

contracts/MarginCalculator.sol
haythem96: To update this, we need to redeploy `MarginCalculator` not the `Controller`, and yeah it will require more txs but I really prefer keeping this constant, as it is more gas efficient, and still if we have failed liquidation w can still batch deployment and params config in one transaction.

contracts/MarginCalculator.sol
haythem96: What do u think is better ? if keeping it conservative will work, I prefer that rather than updating now.

contracts/MarginCalculator.sol
haythem96: It make sense to use the revert for input validation, and false when all input are correct but the vault is not underwater.

contracts/MarginCalculator.sol
haythem96: Why do we need to that in our end ? as we have the function that will return the array of time to expiries, and the function that will return the value for each expiry, so the loop can be done in the side of the dev who need all variables, either in contracts level or fe/web3js as all calls are gasless.

On the other side, I agree with adding a function to set multiple values at once.

antoncoding (dismissed):

contracts/Controller.sol
CruzMolina: @aleone, it looks like the equivalent of `canSync` would be calling `calculator.getExcessCollateral`:

```solidity
(, bool isValidVault) = calculator.getExcessCollateral(vault, typeVault);
```

I think a separate `canSync` function could be useful later on, but don't think it's necessary right now (for example, a FE could poll the vault state for a UI).

contracts/Controller.sol
antoncoding: @haythem96 wanna update this comment?

test/integration-tests/nakedMarginCallPreExpiry.test.ts
```diff
@@ -0,0 +1,776 @@
...
+      optionExpiry = new BigNumber(createValidExpiry(now, 14))
+
+      await otokenFactory.createOtoken(
+        weth.address,
+        usdc.address,
+        weth.address,
+        createTokenAmount(shortStrike, 8),
+        optionExpiry,
+        isPut,
+      )
```
CruzMolina, line 159:
```suggestion
```
I believe this is extra?

CruzMolina: nm, my mistake 😁

antoncoding (approved):

contracts/Controller.sol
```diff
@@ -60,10 +99,17 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
...
+    /// @dev mapping to map vault by each vault type, naked margin vault should be set to 1, spread/max loss vault should be set to 0
+    mapping(address => mapping(uint256 => uint256)) internal vaultType;
+    /// @dev mapping to store the timestamp at which the vault was last updated, will be updated in every action that changes the vault state or when calling sync()
+    mapping(address => mapping(uint256 => uint256)) internal vaultLatestUpdate;
+
     /// @notice emits an event when an account operator is updated for a specific account owner
     event AccountOperatorUpdated(address indexed accountOwner, address indexed operator, bool isSet);
     /// @notice emits an event when a new vault is opened
-    event VaultOpened(address indexed accountOwner, uint256 vaultId);
+    event VaultOpened(address indexed accountOwner, uint256 vaultId, uint256 indexed vaultType);
```
haythem96, line 112: Need to update thegraph later for this, right ? @antoncoding

contracts/Controller.sol
```diff
@@ -379,12 +461,36 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
...
+        external
+        view
+        returns (
+            bool,
+            uint256,
+            uint256
+        )
+    {
+        (, bool isUnderCollat, uint256 price, uint256 dust) = _isLiquidatable(_owner, _vaultId, _roundId);
+        return (isUnderCollat, price, dust);
```
haythem96, line 491: Here I used `isUnderCollat` instead of `isLiquidatable` to avoid the compiler warning re shadowing vars

contracts/Controller.sol
haythem96: done.

test/unit-tests/controllerNakedMargin.test.ts
```diff
@@ -0,0 +1,892 @@
...
+      const liquidatorCollateralBalanceBefore = new BigNumber(await usdc.balanceOf(liquidator))
+
+      await controllerProxy.operate(liquidateArgs, {from: liquidator})
+
+      const liquidatorCollateralBalanceAfter = new BigNumber(await usdc.balanceOf(liquidator))
+      const vaultAfterLiquidation = (await controllerProxy.getVault(accountOwner1, vaultCounter.toString()))[0]
+
+      console.log('vvv aaa lll: ', vaultAfterLiquidation.collateralAmounts[0].toString())
+      console.log('vvv bbb lll: ', new BigNumber(vaultBeforeLiquidation.collateralAmounts[0]).toString())
+      console.log('lll ppp: ', isLiquidatable[1].toString())
```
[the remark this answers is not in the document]
haythem96: 🤣

contracts/Controller.sol
antoncoding: yes, will need to do that. let me create a task so we don't forget

aparnakr (approved):

contracts/pricers/ChainlinkPricer.sol
haythem96: What do u mean by freshness ?

PR #383, opened 2021-04-14
0xCallee to v4 

alexisgauba:
# Task: Update 0xCallee to v4 Feature Name: updated contract and tests

## High Level Description
- Update 0xCallee to work with 0x v4

Specific Changes
- Change order format 
- Change signature format
- Change to batchFillLimitOrders
- Pay protocol fee in ETH

### Todo
- [x] Unit test 100% coverage


.circleci/config.yml
```diff
@@ -25,10 +25,10 @@ workflows:
       - integration-test:
           requires:
             - build
-      - e2e-test:
-          context: COVERALLS_REPO_TOKEN
-          requires:
-            - build
+      # - e2e-test:
```
haythem96, line 28: Why commenting this ?

antoncoding: cause we're no longer using that e2e test (deleted the callee e2e test), only using the testing script now, so want to just disable it. 
I didn't delete the e2e script because I figured that we might want to use it in the future.

antoncoding (approved):

haythem96 (approved):
Good work, LGTM.

PR #385, opened 2021-04-16
Merge setProductTimeToExpiry() & setTimeToExpiryValue() into setUpperBoundValues()

haythem96:
# Task: Merge setProductTimeToExpiry() & setTimeToExpiryValue() into setUpperBoundValues()

## High Level Description

- Remove `setProductTimeToExpiry()` & `setTimeToExpiryValue`
- Add `setUpperBoundValues()` that allow adding many times and p(t) in the same tx
- Add `updateUpperBoundValue()` that allow updating the value for a specific time to expiry

### Code

- [x] Unit test 100% coverage

### Documentation

### DIFF
--- a/contracts/MarginCalculator.sol
+++ b/contracts/MarginCalculator.sol
@@ -106,6 +106,54 @@ contract MarginCalculator is Ownable {
     }
 
     /**
+     * @notice set product upper bound values
+     * @dev can only be called by owner
+     * @param _underlying otoken underlying asset
+     * @param _strike otoken strike asset
+     * @param _collateral otoken collateral asset
+     * @param _isPut otoken type
+     * @param _timesToExpiry array of times to expiry timestamp
+     * @param _values upper bound values array
+     */
+    function setUpperBoundValues(
+        address _underlying,
+        address _strike,
+        address _collateral,
+        bool _isPut,
+        uint256[] calldata _timesToExpiry,
+        uint256[] calldata _values
+    ) external onlyOwner {
+        require(_timesToExpiry.length > 0, "MarginCalculator: invalid times to expiry array");
+        require(_timesToExpiry.length == _values.length, "MarginCalculator: invalid values array");
+
+        // get product hash
+        bytes32 productHash = _getProductHash(_underlying, _strike, _collateral, _isPut);
+        // get array of expiries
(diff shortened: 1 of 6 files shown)


contracts/MarginCalculator.sol
```diff
@@ -106,40 +106,51 @@ contract MarginCalculator is Ownable {
...
         bytes32 productHash = _getProductHash(_underlying, _strike, _collateral, _isPut);
         // get array of expiries
         uint256[] storage expiryArray = productTimeToExpiry[productHash];
 
         // check that this is the first expiry to set, if not the last expiry should be less than the new one to insert (to make sure the array stay in order)
         require(
-            (expiryArray.length == 0) || (_timeToExpiry > expiryArray[expiryArray.length.sub(1)]),
+            (expiryArray.length == 0) || (_timesToExpiry[0] > expiryArray[expiryArray.length.sub(1)]),
             "MarginCalculator: expiry array is not in order"
         );
```
antoncoding, line 138: does this mean it will make it an only "appendable" array?
I feel like maybe we can just restrict it to only be set once. it's unlikely we will want to append some higher `timeToExpiry` value after we first set it. 

or we can just remove this constraint entirely, so we have the privilege to update any value

antoncoding (approved):

aparnakr (approved):

contracts/MarginCalculator.sol
antoncoding: @haythem96 any thoughts?

haythem96: Ah sorry just saw this, the array of times to expiry should always be in ascending order, that how the `_findUpperBoundValue()` look for value.
We may also append higher value after initially setting it.

antoncoding: My question is why do we need this restriction. If it's an admin-only function, why not just open up to set any value in case we make some mistake or want to tune the parameter? removing this can let us reset the whole array. IMO better remove it or just make the setting a **one-time** thing. (cannot reset at all) 

I don't see how setting higher value is useful. the highest timeToExpiry value we will probably set is probably like 6 months til expiry, and we're more likely to adjust closer to expiry timestamp instead of 7 month til expiry in the future.

PR #389, opened 2021-04-23
Improve comments in Calculator & Controller

antoncoding:
# Task: Update comments based on 2 rules:

* avoid explaining what single function do. move those explanation into function description.
* remove some non helpful comments (like add long, increase 1) that's implied by function name.

### Contract updates

1. rename `_price` to `_getDebtPrice` as function description suggested.
2. change expression `if (!(now >= otokenDetails.otokenExpiry)) ` to `if (now < otokenDetails.otokenExpiry))`

### DIFF
--- a/contracts/MarginCalculator.sol
+++ b/contracts/MarginCalculator.sol
@@ -775,7 +775,7 @@ contract MarginCalculator is Ownable {
         // find option upper bound value
         FPI.FixedPointInt memory optionUpperBoundValue = _findUpperBoundValue(_productHash, _shortExpiryTimestamp);
         // convert spot shock value of this product to FixedPointInt (already scaled by 1e27)
-        FPI.FixedPointInt memory spotShockValue = FPI.FixedPointInt(int256(spotShock[_productHash])); // FPI.fromScaledUint(spotShock[_productHash], SCALING_FACTOR);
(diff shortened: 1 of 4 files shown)


contracts/MarginCalculator.sol
```diff
@@ -766,7 +760,7 @@ contract MarginCalculator is Ownable {
         // find option upper bound value
         FPI.FixedPointInt memory optionUpperBoundValue = _findUpperBoundValue(_productHash, _shortExpiryTimestamp);
         // convert spot shock value of this product to FixedPointInt (already scaled by 1e27)
-        FPI.FixedPointInt memory spotShockValue = FPI.fromScaledUint(spotShock[_productHash], SCALING_FACTOR);
+        FPI.FixedPointInt memory spotShockValue = FPI.FixedPointInt(int256(spotShock[_productHash]));
```
antoncoding, line 763: change this because the function `fromScaledUint` was supposed to scale another number that's not in 1e27. 
As the comment suggested, the spotShock value is already in 1e27, so we're safe to just create this FPI struct.

antoncoding: this prevent us from reading the `SCALING_FACTOR` variable

haythem96 (approved):
LGTM.

aparnakr (approved):

PR #392, opened 2021-04-23
Test is Valid Vault + Margin Calculations

aparnakr:
# Task: Test is Valid Vault + Margin Calculations

## High Level Description
Tests the following:
- A lower strike call option should have a greater or equal margin required than a higher strike one ✅
- A lower strike put option should have a less or equal margin required than higher strike one ✅
- A naked margin vault should not have a long otoken

Specific Changes
- Reduced the calls to scaling up, scaling down in getNakedMargin
- To check a vault doesn't have a long otoken, calling _checkIsValidVault is sufficient. No need to go through the math after.


haythem96:
Closing this.

PR #400, opened 2021-05-05
Add public getMarginRequired()

haythem96:
# Task: Add public getMarginRequired()

## High Level Description

Add external getMarginRequired() function.

### Code

- [x] Unit test 100% coverage

### Documentation

### DIFF
--- a/contracts/MarginCalculator.sol
+++ b/contracts/MarginCalculator.sol
@@ -489,18 +489,18 @@ contract MarginCalculator is Ownable {
     }
 
     /**
-     * @notice an external callable _getMarginRequired()
+     * @notice calculate required collateral margin for a vault
      * @param _vault theoretical vault that needs to be checked
-     * @param _vault theoretical vault that needs to be checked
-     * @param _vaultDetails vault details
+     * @param _vaultType vault type
      * @return the vault collateral amount, and marginRequired the minimal amount of collateral needed in a vault, scaled to 1e27
      */
-    function getMarginRequired(MarginVault.Vault memory _vault, VaultDetails memory _vaultDetails)
+    function getMarginRequired(MarginVault.Vault memory _vault, uint256 _vaultType)
         external
         view
         returns (FPI.FixedPointInt memory, FPI.FixedPointInt memory)
     {
-        return _getMarginRequired(_vault, _vaultDetails);
+        VaultDetails memory vaultDetail = _getVaultDetails(_vault, _vaultType);
+        return _getMarginRequired(_vault, vaultDetail);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


contracts/MarginCalculator.sol
```diff
@@ -488,6 +488,21 @@ contract MarginCalculator is Ownable {
         return (true, debtPrice, dust[vaultDetails.shortCollateralAsset]);
     }
 
+    /**
+     * @notice an external callable _getMarginRequired()
```
antoncoding, line 492: Should we use a shorter version of the `_getMarginRequired` description here? instead of saying it's calling another function (kind of meaningless).

CruzMolina: Agree 🙂

haythem96: What other comment u recommend ?

contracts/MarginCalculator.sol
```diff
@@ -488,6 +488,21 @@ contract MarginCalculator is Ownable {
...
+     * @param _vaultDetails vault details
+     * @return the vault collateral amount, and marginRequired the minimal amount of collateral needed in a vault, scaled to 1e27
+     */
+    function getMarginRequired(MarginVault.Vault memory _vault, VaultDetails memory _vaultDetails)
+        external
+        view
+        returns (FPI.FixedPointInt memory, FPI.FixedPointInt memory)
+    {
+        return _getMarginRequired(_vault, _vaultDetails);
+    }
```
antoncoding, line 504:
```suggestion
    function getMarginRequired(MarginVault.Vault memory _vault)
        external
        view
        returns (FPI.FixedPointInt memory, FPI.FixedPointInt memory)
    {
        VaultDetails memory _vaultDetail = _getVaultDetails(_vault);
        return _getMarginRequired(_vault, _vaultDetails);
    }
```

antoncoding (changes requested):
I've been giving some thoughts, I think it make more sense to not asking the user to put in `VaultDetails` as an argument, as the `_getVaultDetails` function is not public. 

In this external function, we should simply get detail first, and that call the internal function.

contracts/MarginCalculator.sol
haythem96: yes make sense, missed that.

contracts/MarginCalculator.sol
antoncoding: I suggest the following: 
`getMarginRequired`: calculate the amount of collateral needed for a vault
`_getMarginRequired`:  calculate the amount of collateral needed for a vault, taking vaultDetail for gas saving

PR #410, opened 2021-05-26
Refactor: VaultSettled & VaultLiquidated events

CruzMolina:
- refactor: index `vaultType` for `VaultSettled` event, scope variables before emitting event to avoid stack too deep error
- docs: update docs for `VaultSettled` change
- refactor: emit `vaultId` for `VaultLiquidated` event, misc style changes
- docs: update docs for `VaultLiquidated` change
- refactor: rm nested scope from `Controller.settleVault`
- refactor: rename `payoutRecipient` -> `to` in `VaultSettled`, `payoutRecipient` -> `receiver` in `VaultLiquidated` events
- docs: update docs

### Code

### Documentation


antoncoding (approved):

haythem96:
let's wait on merging this until we see that we need it.

CruzMolina:
@haythem96 While we no longer need these changes for the v0.1 liq bot, these changes are still helpful for UIs and other developers (same reasoning as before).

antoncoding:
we will need vault id for liquidation event

contracts/core/Controller.sol
```diff
@@ -169,21 +169,23 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
     );
     /// @notice emits an event when a vault is settled
     event VaultSettled(
-        address indexed AccountOwner,
-        address indexed to,
-        address indexed otoken,
+        address indexed accountOwner,
+        address indexed oTokenAddress,
+        address payoutRecipient,
```
haythem96, line 174: Can we keep this to `to` to match the action argument ? IMO better to keep the same naming whenever we can.

contracts/core/Controller.sol
```diff
@@ -169,21 +169,23 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
...
+        uint256 payout,
         uint256 vaultId,
-        uint256 payout
+        uint256 indexed vaultType
     );
     /// @notice emits an event when a vault is liquidated
     event VaultLiquidated(
         address indexed liquidator,
-        address indexed receiver,
+        address indexed payoutRecipient,
```
haythem96, line 182: Same here, can we keep it receiver to match the same name used in the action argument.

contracts/core/Controller.sol
```diff
@@ -899,7 +901,14 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
 
         pool.transferToUser(collateral, _args.to, payout);
 
-        emit VaultSettled(_args.owner, _args.to, address(otoken), _args.vaultId, payout);
+        uint256 vaultId;
+        address payoutRecipient;
+        {
+            vaultId = _args.vaultId;
+            payoutRecipient = _args.to;
+        }
```
haythem96, line 909: I think there is no need for this nested scope, it should work fine by declaring the variable directly outside.

haythem96 (changes requested):
Good work, just left few comments.

antoncoding (approved):
Agree with Haythem's comments

contracts/core/Controller.sol
CruzMolina: Are you open to renaming it to `payoutRecipient`? It's more explicit.

contracts/core/Controller.sol
CruzMolina: Same renaming suggestion here. `payoutRecipient` is much more explicit than `to` :joy:.

haythem96: I agree, but I guess it require some changes in Controller that is not worth it IMO.

contracts/core/Controller.sol
```diff
@@ -169,21 +169,23 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
...
+        uint256 payout,
         uint256 vaultId,
-        uint256 payout
+        uint256 indexed vaultType
     );
     /// @notice emits an event when a vault is liquidated
     event VaultLiquidated(
         address indexed liquidator,
-        address indexed receiver,
+        address indexed to,
```
haythem96, line 182: This called `receiver` in the Action lib

haythem96 (approved):
LGTM! just left one other comment.
For docs, in the future, will be great to just do it in a separate PR.

contracts/core/Controller.sol
CruzMolina: Thanks for catching that!

PR #423, opened 2021-06-03
Build: Setup initial hardhat compilations

CruzMolina:
- build: add `hardhat` as devDep
- build: initial hardhat config w/ solidity settings
- refactor: add `artifacts` & `cache` to `gitignore`
- build: add npm hardhat `compile` script
- refactor: port hardhat config from js to ts
- refactor: rm unused `calculateContractBytecode` script
- build: add `calculate-contract-bytecode` npm script


haythem96:
Why removing `calculateContractBytecode` script ? I sometime use it to check contract bytecode size, I think it is useful.

haythem96:
This is good, I think it will be great if u do a walkthrough the setup or hardhat overall in a learning session before merging this.

CruzMolina:
Ah, my bad @haythem96 . It looked like an unused script. Will re-add.

antoncoding:
Use this instead of the script:
https://hardhat.org/plugins/hardhat-contract-sizer.html

CruzMolina:
@antoncoding the output of the script that plugin runs is different than the `calculateContractBytecode` script.
It outputs a table of each contract's bytecode size in KB but doesn't show how much space is left.

What do you think @haythem96?

Also, IIRC, by default `ganache-cli` maintains the EIP-170 limit. A contract breaching the limit should not deploy locally.

haythem96:
> @antoncoding the output of the script that plugin runs is different than the `calculateContractBytecode` script.
> `hardhat-contract-sizer` outputs a table of each contract's bytecode size in KB but doesn't show how much space is left.
> 
> What do you think @haythem96?
> 
> Also, IIRC, by default `ganache-cli` maintains the EIP-170 limit. A contract breaching the limit should not deploy locally.

Yes that's true, I think it is fine to keep the script, it is already working and doesn't need any maintenance even if we don't mainly use it, so I don't see a cons of keeping it.

CruzMolina:
@haythem96 the issue I see with keeping is that it's very specific to `truffle`'s framework. However, if we decide to remove `truffle` as a devDependency, it shouldn't be difficult to modify the hardhat plugin script to also calculate and provide similar output.

antoncoding:
@CruzMolina can u change the `hardhat.config.ts` file to typescript format? 
Based on the example here: https://hardhat.org/guides/typescript.html
```
export default {
  solidity: "0.7.3",
};
```

antoncoding:
> @haythem96 the issue I see with keeping is that it's very specific to `truffle`'s framework. However, if we decide to remove `truffle` as a devDependency, it shouldn't be difficult to modify the hardhat plugin script to also calculate and provide similar output.

I think we can keep it for now, and open another PR to modify all truffle script to hardhat

haythemsellami:
Do not merge

PR #424, opened 2021-06-05
Build: Setup initial hardhat deployments

CruzMolina:
- style: alphabetize hardhat config keys
- build: add `@nomiclabs/hardhat-ethers` `@nomiclabs/hardhat-waffle` & `ethereum-waffle` as `devDependencies`
- build: configure hardhat config for deployments
- build: port truffle migrations to hardhat deployment script, update npm deployment scripts
- build: use `dotenv/config` for hardhat env vars


hardhat.config.ts
```diff
@@ -1,14 +1,50 @@
...
+  throw new Error("Put mnmenoic into a .secret file, don't paste it in the config.") // deployer wallet mnemonic
+}
+
+infuraApiKey = readFileSync('.infuraKey')
+  .toString()
+  .trim()
+
+mnemonic = readFileSync('.secret')
+  .toString()
+  .trim()
```
antoncoding, line 22: why not use `const insfuraApiKey = ` and `const mnemonic = ` here, but use `let` first and assign later

hardhat.config.ts
```diff
@@ -1,14 +1,50 @@
...
+import '@nomiclabs/hardhat-waffle'
+import {existsSync, readFileSync} from 'fs'
+import {HardhatUserConfig} from 'hardhat/config'
+import {NetworkUserConfig} from 'hardhat/types'
+
+let infuraApiKey: string, mnemonic: string
+
+if (!existsSync('.infuraKey')) {
+  throw new Error("Put infura api key inside a .infuraKey file, don't paste it in the config.") // infura api key
+}
```
antoncoding, line 10: I think it's better to read infura key as env variable (from `.env`), cause we will have etherscan api key, coinmarket cap api key..etc, using a separate file for each one is kind of dumb, we will also need to update gitignore everything we add a new api key.

antoncoding (approved):
Left some comments, glad to see us moving to hardhat. 
Do u want to convert the tests in the `build/hardhat-compile` branch too?

CruzMolina:
@antoncoding Planning to port over the tests... at some point :joy:. For now this PR is intended to switch over migrations to deployments in a non-breaking way.

haythemsellami:
Do not merge

PR #426, opened 2021-06-15
Feat: Oracle migration script

haythem96:
# Feat: Oracle migration script

## High Level Description

This PR add a script to migrate expiry prices from an old oracle module to new Oracle module.
This PR also add `migrateOracle()` in the Oracle module that allow anyone to push prices for an assets.
The function implementation is a bit hacky as it:
- Does not verify if those expiries timestamp are old
- Can set malicious prices if owner is compromised

The above notes are not critical issues IMO, to mitigate any risk:
- deploy Oracle and run the migration script
- Set `migrated` to true by calling `endMigration()`
- Add new Oracle address to AddressBook as the last step

### Code

- [x] Unit test 100% coverage

### Documentation

### DIFF
--- a/scripts/migrateOracle.js
+++ b/scripts/migrateOracle.js
@@ -36,7 +36,7 @@ module.exports = async function(callback) {
       let otokenDetails = await (await Otoken.at(otokenAddress)).getOtokenDetails()
       let expiry = new BigNumber(otokenDetails[4])
 
-      if((otokenDetails[0] == options.asset || otokenDetails[1] == options.asset || otokenDetails[2] == options.asset) && (expiry.toNumber() < currentTimeStamp)) {
+      if((otokenDetails[0] == options.asset || otokenDetails[1] == options.asset || otokenDetails[2] == options.asset) && (expiry.toNumber() < currentTimeStamp) && (!expiriesToMigrate.includes(expiry.toString()))) {
         let priceToMigrate = new BigNumber((await oldOracle.getExpiryPrice(options.asset, expiry.toString()))[0])
 
         expiriesToMigrate.push(expiry.toString())
--- a/scripts/migrateOracle.js
+++ b/scripts/migrateOracle.js
@@ -1,0 +1,59 @@
+const BigNumber = require('bignumber.js')
+
+const yargs = require('yargs')
+
+const OtokenFactory = artifacts.require('OtokenFactory.sol')
+const Oracle = artifacts.require('Oracle.sol')
+const Otoken = artifacts.require('Otoken.sol')
+
+module.exports = async function(callback) {
+  try {
+    const options = yargs
+      .usage('Usage: --network <network> --factory <otokenFactory> --oldOracle <oldOracle> --newOracle <newOracle> --asset <asset> --gasPrice <gasPrice>')
+      .option('network', {describe: '0x exchange address', type: 'string', demandOption: true})
+      .option('factory', {describe: 'Otoken factory address', type: 'string', demandOption: true})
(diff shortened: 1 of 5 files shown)


antoncoding:
why not set an array of expiry and read the value from the old oracle?

haythem96:
> why not set an array of expiry and read the value from the old oracle?

I thought about that, I thought it may consume alot of gas... what do u think ?

antoncoding:
> I thought about that, I thought it may consume alot of gas... what do u think ?

I think it's much lower risk of messing up compared to writing off-chain script to do that. If we only migrate expiry that has otoken created, it's probably just 40~ reads, so it might be okay? 

I don't have really strong preference tbh.

scripts/migrateOracle.js
```diff
@@ -0,0 +1,59 @@
...
+    let expiriesToMigrate = []
+    let pricesToMigrate = []
+
+    for(let i = 0; i < otokenLength.toFixed(); i++) {
+      let otokenAddress = await factory.otokens(i)
+      let otokenDetails = await (await Otoken.at(otokenAddress)).getOtokenDetails()
+      let expiry = new BigNumber(otokenDetails[4])
+
+      if((otokenDetails[0] == options.asset || otokenDetails[1] == options.asset || otokenDetails[2] == options.asset) && (expiry.toNumber() < currentTimeStamp)) {
+        let priceToMigrate = new BigNumber((await oldOracle.getExpiryPrice(options.asset, expiry.toString()))[0])
```
antoncoding, line 40: Why not use `const` for these variables

antoncoding: I think we need a throw an error when the price is 0.

scripts/migrateOracle.js
```diff
@@ -0,0 +1,59 @@
...
+    
+    let expiriesToMigrate = []
+    let pricesToMigrate = []
+
+    for(let i = 0; i < otokenLength.toFixed(); i++) {
+      let otokenAddress = await factory.otokens(i)
+      let otokenDetails = await (await Otoken.at(otokenAddress)).getOtokenDetails()
+      let expiry = new BigNumber(otokenDetails[4])
+
+      if((otokenDetails[0] == options.asset || otokenDetails[1] == options.asset || otokenDetails[2] == options.asset) && (expiry.toNumber() < currentTimeStamp)) {
```
antoncoding, line 39: we can also include expiry that's already be set right? Maybe use a Set to keep track of all the expiry that's already pushed into the `expiriesToMigrate` array

haythem96: Added that

scripts/migrateOracle.js
haythem96: let's skip this for now, as we know when we run the script that all expired options already have a pushed price.

scripts/migrateOracle.js
```diff
@@ -0,0 +1,59 @@
...
+    const currentTimeStamp = Math.floor(Date.now() / 1000);
+    const otokenLength = new BigNumber(await factory.getOtokensLength())
+    
+    let expiriesToMigrate = []
+    let pricesToMigrate = []
+
+    for(let i = 0; i < otokenLength.toFixed(); i++) {
+      let otokenAddress = await factory.otokens(i)
+      let otokenDetails = await (await Otoken.at(otokenAddress)).getOtokenDetails()
+      let expiry = new BigNumber(otokenDetails[4])
```
antoncoding, line 37: Can you use `const` instead of `let`?

antoncoding (approved):

aparnakr (approved):

PR #428, opened 2021-06-18
Feat: Naked margin cap across all vaults

haythem96:
# Feat: Naked margin cap across all vaults

## High Level Description

This PR change the naked margin cap from per vault to a cap across all collateral vaults in pool (naked margin vaults).
To implement the 🧢  across all vaults, a mapping `nakedPoolBalance` track all collateral deposited, withdrawn, settled and liquidated from a naked margin vault. The `nakedPoolBalance` for a specific collateral asset should never be greater than `nakedCap`.

### Code

- [x] Unit test 100% coverage

### Documentation


haythem96:
The check if naked collateral across all vaults is greater than the naked cap is executed in the `_depositCollateral()` https://github.com/opynfinance/GammaProtocol/pull/428/files#diff-691d8bb3951fcb892aacab5de51e61e4a0d472907e30018c0987c5828f788c40R771

In the scenario where the naked cap for WETH is 500k, and there is 400k deposited as collateral, some vault are underwater and a liquidation is needed, if liquidator send an array of actions [open vault, deposit collateral, mint short, liquidate] where the collateral > 100k WETH, that will revert. It is better to always liquidate and then deposit collateral if both those actions are needed.

contracts/core/Controller.sol
```diff
@@ -945,6 +985,9 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
         // decrease amount of short otoken in liquidated vault, index of short otoken to decrease is hardcoded at 0
         vaults[_args.owner][_args.vaultId].removeShort(vault.shortOtokens[0], _args.amount, 0);
 
+        // decrease internal naked margin collateral amount
+        nakedPoolBalance[vault.collateralAssets[0]] = nakedPoolBalance[vault.collateralAssets[0]].sub(collateralToSell);
+
```
antoncoding, line 990: this should only happen if vaultType = 1 right?

antoncoding (approved):
LGTM

contracts/core/Controller.sol
antoncoding: never mind, this is under liquidation 😄

haythem96: Yes, _liquidate function will only work if vaultType is 1 anyway as this will revert if it is not https://github.com/opynfinance/GammaProtocol/blob/3ee964999ff7642251d1db3efeb4dd2336d5c59d/contracts/core/Controller.sol#L961

contracts/core/Controller.sol
```diff
@@ -56,6 +56,8 @@ import {CalleeInterface} from "../interfaces/CalleeInterface.sol";
  * CO33: can not liquidate vault
  * CO34: can not leave less than collateral dust
  * CO35: invalid vault id
+ * CO36: cap amount should be greater than zero
+ * CO37: collateral exceed naked margin cap
```
CruzMolina, line 60:
```suggestion
 * CO37: collateral exceeds naked margin cap
```

aparnakr (approved):

CruzMolina (approved):
LGTM

PR #437, opened 2021-06-29
Chore: Refactor oracle migration script to use subgraph

antoncoding:
# Task: Update oracle migration script to fetch price from subgraph

* replace loop of web3 calls with 1 subgraph query.
* this script only work under the assumption that the subgraph have correct price history


haythem96:
What's the status here ? @antoncoding

antoncoding:
> What's the status here ? @antoncoding

updated

aparnakr (approved):

PR #441, opened 2021-08-18
Feat: StakedaoEcrvPricer

CruzMolina:
- feat: initial `ICurve` contract
- feat: initial `IStakeDao` contract
- feat: initial `StakedaoEcrvPricer` contract
- feat: initial `StakedaoEcrvPricer` deployment script
- docs: update `README` w/ `StakedaoEcrvPricer` deployment steps
- docs: update docs & diagrams


contracts/interfaces/ICurve.sol
```diff
@@ -0,0 +1,16 @@
...
+pragma experimental ABIEncoderV2;
+
+interface ICurve {
+    function add_liquidity(uint256[2] memory amounts, uint256 minAmount) external payable returns (uint256);
+
+    function remove_liquidity_one_coin(
+        uint256 _token_amount,
+        int128 i,
+        uint256 _minAmount
+    ) external returns (uint256);
```
haythem96, line 13: I don't think we need those functions.

contracts/interfaces/IStakeDao.sol
```diff
@@ -0,0 +1,26 @@
...
+
+    function withdraw(uint256 _shares) external;
+
+    function token() external returns (ERC20Interface);
+
+    function balanceOf(address account) external view returns (uint256);
+
+    function approve(address spender, uint256 amount) external returns (bool);
+
+    function transfer(address recipient, uint256 amount) external returns (bool);
```
haythem96, line 23: Same here.

contracts/interfaces/IStakeDao.sol
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: MIT
+
+import {ERC20Interface} from "./ERC20Interface.sol";
+
+pragma solidity 0.6.10;
+pragma experimental ABIEncoderV2;
```
haythem96, line 6: Why adding this line here ?

contracts/interfaces/ICurve.sol
haythem96: Why using `ABIEncoderV2` in this file ?

CruzMolina: This contract is from the StakeDao templates. I don't think it's a good idea to have a different contract here than what's in the `stakedao-master` branch in the perp vault templates repo.

contracts/interfaces/IStakeDao.sol
CruzMolina: This contract is also from the StakeDao templates. I don't think it's a good idea to have a different contract here than what's in the `stakedao-master` branch in the perp vault templates repo.

contracts/interfaces/ICurve.sol
haythem96: I guess we are always only exposing functions that we use in interfaces, it is bit confusing to see all of them and it mislead that we use those too also.

contracts/pricers/StakedaoEcrvPricer.sol
```diff
@@ -0,0 +1,104 @@
...
+ * @notice A Pricer contract for a Stakedao lpToken
+ */
+contract StakedaoEcrvPricer {
+    using SafeMath for uint256;
+
+    /// @notice curve pool
+    ICurve public curve;
+
+    /// @notice underlying asset for this lpToken
+    ERC20Interface public underlying;
```
haythem96, line 34: I think we can store this as `address`

contracts/pricers/StakedaoEcrvPricer.sol
```diff
@@ -0,0 +1,104 @@
...
+     * @dev convert underlying price to lpToken price with the lpToken to underlying exchange rate
+     * @param _underlyingPrice price of 1 underlying token (ie 1e6 USDC, 1e18 WETH) in USD, scaled by 1e8
+     * @return price of 1e8 lpToken in USD, scaled by 1e8
+     */
+    function _underlyingPriceToYtokenPrice(uint256 _underlyingPrice) private view returns (uint256) {
+        uint256 pricePerShare = lpToken.getPricePerFullShare();
+        uint8 underlyingDecimals = 18;
+        uint256 curvePrice = curve.get_virtual_price();
+
+        return pricePerShare.mul(_underlyingPrice).mul(curvePrice).div(10**uint256(2 * underlyingDecimals));
```
haythem96, line 98: I think we don't need to store `18` in `uint8` and then convert it back to `uint256`. IMO will be better to just write `.div(1e36)` and explain from where the `36` came from.

haythem96: Also what is the lpToken and the underlying that will be used in this pricer ? is the underlying always with 18 decimals ?

contracts/interfaces/ICurve.sol
CruzMolina: Yeah I agree. It would be nice to have a separate repo for the pricers (or have this be a monorepo) and have the pricer contracts packaged and imported for deployment to minimize redundancy.

contracts/pricers/StakedaoEcrvPricer.sol
CruzMolina: The `lpToken` currently is [`sdeCRV`](https://etherscan.io/address/0xa2761b0539374eb7af2155f76eb09864af075250). For the current deployed pricer, the underlying is `wETH`, which is why 18 is hardcoded. Would need to change it if underlying is changed.

CruzMolina: Updated code comments for this.

contracts/interfaces/ICurve.sol
aparnakr: I'm ok keeping or removing. It def is cleaner to remove.

aparnakr (approved):

aparnakr (approved):

PR #444, opened 2021-11-16
feat: pricer v2.0

haythem96:
# Task: pricer v2.0

## High Level Description

Specific Changes
Function x was added ...

### Code

- [ ] Unit test 100% coverage

### Documentation

### DIFF
--- a/contracts/pricers/ChainlinkPricer.sol
+++ b/contracts/pricers/ChainlinkPricer.sol
@@ -25,26 +25,20 @@ contract ChainLinkPricer is OpynPricerInterface {
 
     /// @notice asset that this pricer will a get price for
     address public asset;
-    /// @notice bot address that is allowed to call setExpiryPriceInOracle
-    address public bot;
 
     /**
-     * @param _bot priveleged address that can call setExpiryPriceInOracle
      * @param _asset asset that this pricer will get a price for
      * @param _aggregator Chainlink aggregator contract for the asset
      * @param _oracle Opyn Oracle address
      */
     constructor(
-        address _bot,
         address _asset,
         address _aggregator,
         address _oracle
     ) public {
-        require(_bot != address(0), "ChainLinkPricer: Cannot set 0 address as bot");
         require(_oracle != address(0), "ChainLinkPricer: Cannot set 0 address as oracle");
         require(_aggregator != address(0), "ChainLinkPricer: Cannot set 0 address as aggregator");
 
-        bot = _bot;
         oracle = OracleInterface(_oracle);
         aggregator = AggregatorInterface(_aggregator);
         asset = _asset;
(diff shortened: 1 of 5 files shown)


contracts/pricers/ChainlinkPricer.sol
```diff
@@ -25,53 +25,52 @@ contract ChainLinkPricer is OpynPricerInterface {
...
+        while (!isCorrectRoundId) {
+            (, int256 price, , uint256 roundTimestamp, ) = aggregator.getRoundData(previousRoundId);
+
+            if (roundTimestamp == 0) {
+                previousRoundId--;
+            } else if (roundTimestamp > _expiryTimestamp) {
+                revert("ChainLinkPricer: invalid roundId");
+            } else {
+                isCorrectRoundId = true;
+            }
```
aparnakr, line 71: just trying to understand, is there a chance this could revert and leave us in a position where no one could push prices at all? (Like what assumptions about chainlink data can we make?)

aparnakr (approved):

contracts/pricers/ChainlinkPricer.sol
haythem96: Yeah if this `while loop` run out of gas, like there is 1000 gap between previous ID and current one, we need to push price by our ownable function.

antoncoding:
What's the benefit of updating the oracle right now?

functionally, isn't it the same cuz we are already able to dispute an old price? why do we need this oracle update right now with the new pricer updates.

I don't think we have time to run oracle migration right now, so not sure why we're updating the Oracle contract

antoncoding:
Can we just implement a function that allow us to set price from the pricer? (literally the exact old function)

Just so we can limit the scope to the pricer, avoid the migration

test/unit-tests/chainlinkPricer.test.ts
```diff
@@ -100,6 +92,7 @@ contract('ChainlinkPricer', ([owner, bot, random]) => {
     before('setup history in aggregator', async () => {
       // set t0, t1, t2, expiry, t3, t4
       t0 = (await time.latest()).toNumber()
+      console.log('t0', t0)
```
antoncoding, line 95: remove

haythem96:
> Can we just implement a function that allow us to set price from the pricer? (literally the exact old function)
> 
> Just so we can limit the scope to the pricer, avoid the migration. (a full migration will require us to redeploy all the pricers, and corresponding infra) I really think just adding a public function solve 95% of the problem while maintaining the same level of centralization power.

Yes I agree, but don't you think it is not the perfect design and way of doing things ?! just adding duplicate functions ?

antoncoding:
the benefit is to keep the complexity only in the Pricer, and won't make our role as Oracle Admin too powerful. 
(with this design, admin can force to write a settlemet price first and disable pricers to override it, which is more centralized)

PR #459, opened 2022-07-19
Add Borrowable Margin Pool (V2)

chuddster:
# Task: Feature Name:

## High Level Description

Allow whitelisted ribbon vaults to create a short position and direct their collateral to the new borrowable margin pool (MarginPoolV2) instead of the unborrowable marginpool (MarginPool). This allows certain whitelisted parties (whitelisted directly on MarginPoolV2) to borrow the collateral backing the call option. The whitelisted parties will exclusively be market makers who also buy the call option. This allows the market maker to delta hedge in a more capital efficient manner since they have an uncollateralized loan to the spot asset (which they can then short). We will whitelist vaults of treasuries that agree to offer this uncollateralized loan in exchange for higher premiums on the call option. 

Say Zebra DAO has a $ZBR token. They launch a treasury vault with ribbon finance and deposit 100 $ZBR to sell covered calls and earn premiums from market makers. The 100 $ZBR is deposited into MarginPoolV2 when oTokens are minted. Market Maker X has 10 ZBR oTokens (call options) that they bought from the vault's oToken auction. Whereas in vanilla MarginPool the collateral cannot be taken out until the end, in MarginPoolV2 the oToken buyer (market maker) can borrow 10 $ZBR from MarginPoolV2 using borrow() to hedge their positive delta, posting **ZERO** collateral. Notice the amount they can borrow is proportional to how many oTokens they bought. Once a month passes, they deposit 10 $ZBR back into the vault.

Controller.sol: 
- setMarginPoolV2: adds margin pool v2 address
- donateV2: donates to v2 margin pool

MarginPool.sol
- setBorrowerWhitelistedStatus: sets whether an address is whitelisted to borrow
- setRibbonVaultWhitelistedStatus: sets whether ribbon vault address is whitelisted to create short oToken on MarginPoolV2
- setBuyerWhitelistedStatus: sets whether address is whitelisted as oToken buyer of oToken minted on MarginPoolV2 
- setBorrowPCT: sets the borrow percentage of entire collateral on a per-asset basis
- borrow: allows whitelisted address (market maker) to borrow some of underlying collateral proportional to their oToken bal
- borrowable: checks how much an address can borrow of underlying
- repay: allows whitelisted address to repay their loan once oToken expires
- repayFor: allows any address to repay loan on behalf of another party

### Code

- [ ] Unit test 100% coverage

### Documentation


contracts/core/MarginPoolV2.sol
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    /**
+     * @notice Lends out asset to market maker
+     * @param _oToken address of the oToken laying claim to the collateral assets
+     * @param _amount amount of the asset to lend to borrower
+     */
+    function borrow(address _oToken, uint256 _amount) public onlyWhitelistedBorrower {
+        require(
+            WhitelistInterface(AddressBookInterface(addressBook).getWhitelist()).isWhitelistedOtoken(_oToken),
```
chuddster, line 125: Check the Whitelist if the otoken is created. This means they have to create the otokens via the factory and prevents fake otokens to drain some underlying

contracts/core/Controller.sol
```diff
@@ -910,7 +909,12 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
 
         otoken.burnOtoken(msg.sender, _args.amount);
 
-        pool.transferToUser(collateral, _args.receiver, payout);
+        MarginPoolInterface _pool = (!poolV2.whitelistedOTokenBuyer(msg.sender) ||
+            ERC20Interface(collateral).balanceOf(address(poolV2)) < payout)
+            ? pool
+            : poolV2;
+
```
chuddster, line 916: logic for _pool: in redeem() we have no reference to a vault where we can get the margin pool. This method just redeems collateral and the only given is the oToken address (no margin pool identification). So the way to guess the right margin pool to redeem from is to check if oToken owner is whitelisted in our ad-hoc mapping in pool v2. If it is whitelisted, we know for sure it is a market maker. But, it is still possible for market maker to hold oTokens for non-treasury vaults - in which case only first condition would be insufficient. We then check if the balance of that asset, like ETH, in the new margin pool is less than the payout. If it is then we know they are holding a normal otoken from say the ETH retail vault. Unless someone donates ETH for free to exclusively treasury vault margin pool, this will correctly identify which margin pool to retrieve from.

contracts/core/Controller.sol
```diff
@@ -719,6 +707,9 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
         // store new vault
         accountVaultCounter[_args.owner] = vaultId;
         vaultType[_args.owner][vaultId] = _args.vaultType;
+        vaults[_args.owner][vaultId].setMarginPool(
+            poolV2.whitelistedRibbonVault(_args.owner) ? address(poolV2) : address(pool)
+        );
```
chuddster, line 713: when a vault is created, we immediately associate it with a margin pool in the [Vault](https://github.com/ribbon-finance/avocado/blob/master/contracts/libs/MarginVault.sol#L33) struct. We determine which margin pool to set by checking if owner of vault is a whitelisted address from the margin pool v2 ad-hoc mapping. Addresses of balancer, badger, spell vaults (and others) will be whitelisted. Retail vaults will not be and as a result will be directed to vanilla margin pool like before

contracts/core/Controller.sol
```diff
@@ -458,26 +466,6 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
         return operators[_owner][_operator];
     }
 
-    /**
```
chuddster, line 461: save gas

contracts/core/Controller.sol
```diff
@@ -381,8 +384,6 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
      * @param _isRestricted new call restriction state
      */
     function setCallRestriction(bool _isRestricted) external onlyOwner {
-        require(callRestricted != _isRestricted, "C9");
```
chuddster, line 384: save gas

contracts/core/MarginPoolV2.sol
```diff
@@ -0,0 +1,244 @@
...
+    mapping(address => uint256) public borrowPCT;
+    /// @dev mapping between address and whitelist status of borrower
+    mapping(address => bool) public whitelistedBorrower;
+    /// @dev mapping between address and whitelist status of otoken buyer
+    /// This is the whitelist for all the holders of oTokens that have a claim to
+    /// collateral in this pool
+    mapping(address => bool) public whitelistedOTokenBuyer;
+    /// @dev mapping between address and whitelist status of ribbon vault
+    mapping(address => bool) public whitelistedRibbonVault;
+    /// @dev mapping between (borrower, asset) and outstanding borrower amount
```
chuddster, line 35: mappings that are crucial for knowing whether to direct vaults to margin pool v1 or v2

whitelistedOTokenBuyer: knowing which margin pool to claim an oToken from
whitelistedRibbonVault: knowing which margin pool to set a margin vault to

contracts/core/Controller.sol
```diff
@@ -74,6 +72,7 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
     OracleInterface public oracle;
     MarginCalculatorInterface public calculator;
     MarginPoolInterface public pool;
+    MarginPoolInterface public poolV2;
```
haythem96, line 75: Suggest renaming this to `borrowablePool`

contracts/core/Controller.sol
```diff
@@ -324,14 +323,24 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
...
+        emit Donated(msg.sender, _asset, _amount);
+    }
+
     /**
      * @notice allows the partialPauser to toggle the systemPartiallyPaused variable and partially pause or partially unpause the system
      * @dev can only be called by the partialPauser
      * @param _partiallyPaused new boolean value to set systemPartiallyPaused to
      */
     function setSystemPartiallyPaused(bool _partiallyPaused) external onlyPartialPauser {
-        require(systemPartiallyPaused != _partiallyPaused, "C9");
```
haythem96, line 333: IMO no harm for keeping those and it make it feel safer, also the owner is the only one capable of calling this function so it is fine to pay for extra gas

contracts/core/Controller.sol
```diff
@@ -343,8 +352,6 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
      * @param _fullyPaused new boolean value to set systemFullyPaused to
      */
     function setSystemFullyPaused(bool _fullyPaused) external onlyFullPauser {
-        require(systemFullyPaused != _fullyPaused, "C9");
```
haythem96, line 346: Similar comment to the above require statement

contracts/core/Controller.sol
```diff
@@ -356,8 +363,6 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
      * @param _fullPauser new fullPauser address
      */
     function setFullPauser(address _fullPauser) external onlyOwner {
-        require(_fullPauser != address(0), "C10");
-        require(fullPauser != _fullPauser, "C9");
```
haythem96, line 360: Similar comment to the above require statement

contracts/core/Controller.sol
haythem96: Added comment about removing the require above

contracts/core/Controller.sol
```diff
@@ -395,8 +396,6 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
      * @param _isOperator new boolean value that expresses if the sender is giving or revoking privileges for _operator
      */
     function setOperator(address _operator, bool _isOperator) external {
-        require(operators[msg.sender][_operator] != _isOperator, "C9");
```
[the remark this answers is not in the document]
haythem96: Agee on removing here.

contracts/core/Controller.sol
```diff
@@ -423,6 +422,15 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
...
 
+    /**
+     * @notice Sets margin pool v2
+     * @dev can only be called by the owner
+     * @param _marginPoolV2 margin pool v2 address
+     */
+    function setMarginPoolV2(address _marginPoolV2) external onlyOwner {
+        poolV2 = MarginPoolInterface(_marginPoolV2);
+    }
+
```
haythem96, line 433: This one can be removed, the `borrowablePool` address should be set in the AddressBook (following the current architecture) using the `AddressBook.setAddress()` function, then the address can be fetched and assigned to this variable in the `_refreshConfigInternal()`

```
borrowablePool = MarginPoolInterface(addressbook. getAddress(keccak256"BORROWABLE_POOL"));
```

contracts/core/Controller.sol
```diff
@@ -458,26 +466,6 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
...
     }
 
-    /**
-     * @notice returns the current controller configuration
-     * @return whitelist, the address of the whitelist module
-     * @return oracle, the address of the oracle module
-     * @return calculator, the address of the calculator module
-     * @return pool, the address of the pool module
-     */
-    function getConfiguration()
```
haythem96, line 468: Why removing this one?

contracts/libs/MarginVault.sol
```diff
@@ -31,6 +31,8 @@ library MarginVault {
 
     // vault is a struct of 6 arrays that describe a position a user has, a user can have multiple vaults.
     struct Vault {
+        // address of the margin pool for vault
```
haythem96, line 34: IMO it is better to keep using vaultType instead of adding address pool in the vault structure, currently we have 2 vault types, we can use `3` as borrowable vault, and then do the checks if vault type is 3 when depositing collateral, forward it to the borrowbalePool

haythem96 (changes requested):
IMO better to remove the margin pool address from the vault data struct and use a new vault type `3` for borrowable vault, and if a vault is type 3, then forward the collateral to deposit to borrowablePool (basically vault type 3 interact with the new pool)

contracts/core/Controller.sol
```diff
@@ -719,6 +707,9 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
         // store new vault
         accountVaultCounter[_args.owner] = vaultId;
         vaultType[_args.owner][vaultId] = _args.vaultType;
+        vaults[_args.owner][vaultId].setMarginPool(
+            poolV2.whitelistedRibbonVault(_args.owner) ? address(poolV2) : address(pool)
```
haythem96, line 711: I guess we can rename this check function to `isWhitelistedBorrowableVault()`

contracts/core/MarginPoolV2.sol
```diff
@@ -0,0 +1,244 @@
...
+    /// @dev mapping between collateral asset and borrow PCT
+    mapping(address => uint256) public borrowPCT;
+    /// @dev mapping between address and whitelist status of borrower
+    mapping(address => bool) public whitelistedBorrower;
+    /// @dev mapping between address and whitelist status of otoken buyer
+    /// This is the whitelist for all the holders of oTokens that have a claim to
+    /// collateral in this pool
+    mapping(address => bool) public whitelistedOTokenBuyer;
+    /// @dev mapping between address and whitelist status of ribbon vault
+    mapping(address => bool) public whitelistedRibbonVault;
```
haythem96, line 34: Rename to a generic naming `whitelistedBorrowableVault`

contracts/libs/MarginVault.sol
antoncoding: changing the vault struct will mess up with all existing vault in the storage right? or is it safe because all existing properties are arrays so they're not stored based on order?

contracts/core/MarginPoolV2.sol
```diff
@@ -0,0 +1,244 @@
...
+        );
+
+        OtokenInterface oToken = OtokenInterface(_oToken);
+
+        address collateralAsset = oToken.collateralAsset();
+        uint256 outstandingAssetBorrow = borrowed[msg.sender][collateralAsset];
+
+        require(!oToken.isPut(), "MarginPool: oToken is not a call option");
+
+        require(borrowPCT[collateralAsset] > 0, "MarginPool: Borrowing is paused for collateral asset");
```
antoncoding, line 136: i think this is a duplicate check with L155 
```
require(_amount <= collateralAllocatedToMM.mul(borrowPCT[collateralAsset]).div(TOTAL_PCT).sub(outstandingAssetBorrow),
            "MarginPool: Borrowing more than allocated"
        );
```
Can remove to save gas

contracts/core/MarginPoolV2.sol
```diff
@@ -0,0 +1,244 @@
...
+            _amount <=
+                collateralAllocatedToMM.mul(borrowPCT[collateralAsset]).div(TOTAL_PCT).sub(outstandingAssetBorrow),
+            "MarginPool: Borrowing more than allocated"
+        );
+
+        borrowed[msg.sender][collateralAsset] = outstandingAssetBorrow.add(_amount);
+
+        // transfer _asset _amount from pool to borrower
+        ERC20Interface(collateralAsset).safeTransfer(msg.sender, _amount);
+        emit Borrow(_oToken, collateralAsset, _amount, msg.sender);
```
antoncoding, line 165: Two tricks that would probably mess up the accounting: 
1. whitelisted parties can collude and make others unable to borrow: they can transfer the oToken balance back and forth among whitelisted parties and borrow more than what they should have.
2. whitelisted parties can flash mint bunch of oTokens and borrow almost every borrowable asset from the pool.

I think the pool should lock up oToken when someone come and borrow, to prevent these scenarios

contracts/core/MarginPoolV2.sol
```diff
@@ -0,0 +1,244 @@
...
+import {SafeMath} from "../packages/oz/SafeMath.sol";
+import {SafeERC20} from "../packages/oz/SafeERC20.sol";
+import {MarginPool} from "./MarginPool.sol";
+
+/**
+ * @author Ribbon Team
+ * @title MarginPoolV2
+ * @notice Contract that holds all protocol funds AND allows collateral borrows
+ */
+contract MarginPoolV2 is MarginPool {
```
haythem96, line 19: `BorrowableMarginPool` make more sense as naming than `MarginPoolV2` IMO

contracts/core/MarginPoolV2.sol
haythem96: Agree

contracts/core/Controller.sol
chuddster: I would have left it, but otherwise I run into contract size issues and this seemed like the least consequential to remove.

contracts/core/Controller.sol
chuddster: Ah, I didn't realize this was possible. Good solution!

contracts/core/Controller.sol
chuddster: contract size

contracts/libs/MarginVault.sol
chuddster: Yes I will set vault type index 2 to borrowable pool. Upon second glance I realize this won't cause any problems with payoff calculation / liquidation logic in MarginCalculator because this pool will only be used for covered calls (not a naked vault) so the pool and naked vaults are mutually exclusive.

contracts/core/MarginPoolV2.sol
chuddster: good catch

contracts/core/MarginPoolV2.sol
chuddster: Scenario #1 is possible. I just chalked it up to "master loan agreement will be signed so there is no incentive here" but locking up oTokens definitely solves this. 

Scenario #2 flashminting is solved because whereas before they could flash mint whitelisted oToken and borrow underlying from pool and then just burn the original oTokens - with a lock they are basically swapping their own collateral with that of the pools so there is no gain here.

chuddster: @antoncoding @haythem96 thanks for the thorough review

contracts/core/Controller.sol
haythem96: Okay this was removed to avoid contract bytecode size, revisit later....

contracts/core/Controller.sol
```diff
@@ -368,8 +373,6 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
      * @param _partialPauser new partialPauser address
      */
     function setPartialPauser(address _partialPauser) external onlyOwner {
-        require(_partialPauser != address(0), "C11");
-        require(partialPauser != _partialPauser, "C9");
```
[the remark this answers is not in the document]
haythem96: Okay this was removed to avoid contract bytecode size, revisit later...

contracts/core/BorrowableMarginPool.sol
```diff
@@ -0,0 +1,288 @@
...
+    /// @dev mapping between address and whitelist status of borrower
+    mapping(address => bool) internal whitelistedBorrower;
+    /// @dev mapping between address and whitelist status of otoken buyer
+    /// This is the whitelist for all the holders of oTokens that have a claim to
+    /// collateral in this pool
+    mapping(address => bool) internal whitelistedOTokenBuyer;
+    /// @dev mapping between address and whitelist status of options vault
+    /// This denotes whether an options vault can create a margin vault with
+    /// collateral custodied in this borrowable vault
+    mapping(address => bool) internal whitelistedOptionsVault;
```
haythem96, line 36: Sorry if it seems there are alot of naming comments, but let's try to use `borrowableVault` and `borrowablePool` terminology everywhere, so make sense here IMO to rename it to `whitelistedBorrowableVault`, as also vault type 2 indicate a borrowable vault.

contracts/core/BorrowableMarginPool.sol
```diff
@@ -0,0 +1,288 @@
...
+     * @param _oTokenBuyer address of the oToken buyer
+     * @param _whitelisted bool of whether it is whitelisted or blacklisted
+     */
+    function setOTokenBuyerWhitelistedStatus(address _oTokenBuyer, bool _whitelisted) external onlyOwner {
+        require(_oTokenBuyer != address(0), "!_buyer");
+
+        whitelistedOTokenBuyer[_oTokenBuyer] = _whitelisted;
+        emit SetOTokenBuyerWhitelist(_oTokenBuyer, _whitelisted);
+    }
+
```
haythem96, line 136: Just thinking loud here, if it make sense to keep this here or move it somehow to the Whitelist module

contracts/core/BorrowableMarginPool.sol
```diff
@@ -0,0 +1,288 @@
...
+            "MarginPool: Cannot borrow collateral asset of expired oToken"
+        );
+
+        uint256 collateralAssetDecimals = ERC20Interface(collateralAsset).decimals();
+
+        uint256 oTokenBalance = ERC20Interface(_oToken).balanceOf(msg.sender);
+
+        // Each oToken represents 1:1 of collateral token
+        // So a market maker can borrow at most our oToken balance in the collateral asset
+        uint256 collateralAllocatedToMM = collateralAssetDecimals >= 8
```
haythem96, line 179: `collateralAllocatedToMM` => `collateralAllocatedToBorrower`

contracts/core/BorrowableMarginPool.sol
```diff
@@ -0,0 +1,288 @@
...
+            ? oTokenBalance.mul(10**(collateralAssetDecimals.sub(8)))
+            : oTokenBalance.div(10**(uint256(8).sub(collateralAssetDecimals)));
+
```
haythem96, line 182: Can u explain more the >= 8 decimals ? maybe also add some inline comments for it

contracts/core/BorrowableMarginPool.sol
```diff
@@ -0,0 +1,288 @@
...
+        borrowPCT[_collateral] = _borrowPCT;
+        emit SetBorrowPCT(_collateral, _borrowPCT);
+    }
+
+    /**
+     * @notice Lends out asset to market maker
+     * @param _oToken address of the oToken laying claim to the collateral assets
+     * @param _amount amount of the asset to lend to borrower
+     */
+    function borrow(address _oToken, uint256 _amount) public onlyWhitelistedBorrower {
```
haythem96, line 152: I guess here it make more sense to allow borrower to set the amount of oToken they are giving, and calculate amount of collateral asset to borrow, instead of doing it the other way, also allow for borrower to borrow and keep some oToken as this is using all the msg.sender oToken balance.

contracts/core/BorrowableMarginPool.sol
```diff
@@ -0,0 +1,288 @@
...
+    /**
+     * @notice Repays asset back to pool for another borrower before oToken expiry
+     * @param _oToken address of the oToken laying claim to the collateral assets
+     * @param _amount amount of the asset to repay
+     * @param _borrower address of the borrower to repay for
+     * @param _repayer address of the repayer of the loan
+     */
+    function _repay(
+        address _oToken,
+        uint256 _amount,
```
haythem96, line 263: `_amount` is confusing between oToken amount of which amount, IMO better to name it `_collateralAmount`

contracts/core/Controller.sol
```diff
@@ -324,14 +324,24 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
         emit Donated(msg.sender, _asset, _amount);
     }
 
+    /**
+     * @notice send asset amount to margin pool v2
+     * @dev use donate() instead of direct transfer() to store the balance in assetBalance
+     * @param _asset asset address
+     * @param _amount amount to donate to pool
+     */
+    function donateV2(address _asset, uint256 _amount) external {
```
haythem96, line 333: `donateBorrowablePool()`

contracts/core/Controller.sol
```diff
@@ -718,7 +698,7 @@ contract Controller is Initializable, OwnableUpgradeSafe, ReentrancyGuardUpgrade
 
         // store new vault
         accountVaultCounter[_args.owner] = vaultId;
-        vaultType[_args.owner][vaultId] = _args.vaultType;
+        vaultType[_args.owner][vaultId] = borrowablePool.isWhitelistedOptionsVault(_args.owner) ? 2 : _args.vaultType;
```
haythem96, line 701: I think here better to keep it as the original implementation, and call `borrowablePool.isWhitelistedOptionsVault()` to check if whitelisted only if `_args.vaultType == 2`, it is more flexible as it allow vault owner even when whitelisted to open naked margin/spread vault, and also make the external call to check if whitelisted only if type==2

contracts/core/Controller.sol
haythem96: MM I'm not sure this is safe way to do it, in the example of MM holding oTokens for non-treasury vaults + treasury vaults, and they want to redeem the non-treasury one, and assuming the borrowable pool have enough balance, the amount will be deducted from the borrowable pool where it should be withdrawn from the normal pool, right?

haythem96: Say `MM1` have 500oToken, and 400oToken(belong to borrowable pool), and borrowable pool have 600USDC(for the 400oToken and other borrowers MM2,3...etc) MM1 come to redeem 500oToken, which should be redeemed from the normal pool, it will be redeemed from the borrowable pool

haythem96: Also is there a scenario where it is more profitable for the borrower to not repay?

haythem96: I feel I'm more towards having one margin pool now, with a mapping of borrowable amounts...

haythem96 (commented):
Left some comments, main concerns re figuring out which pool to redeem oToken from. I actually think upgrading the MarginPool and having one MarginPool that keep track of collateral that can be borrowed may be better

contracts/core/BorrowableMarginPool.sol
haythem96: Yeah I think moving this to Whitelist.sol make sense

haythem96:
> Left some comments, main concerns re figuring out which pool to redeem oToken from. I actually think upgrading the MarginPool and having one MarginPool that keep track of collateral that can be borrowed may be better

Also probably in that case whitelisting oToken buyer is not needed...

haythem96:
> > Left some comments, main concerns re figuring out which pool to redeem oToken from. I actually think upgrading the MarginPool and having one MarginPool that keep track of collateral that can be borrowed may be better
> 
> Also probably in that case whitelisting oToken buyer is not needed...

Something like 1 pool contract (but we need to migrate current pool funds to the new one, but can be done by adding a function in Controller that can only be called by owner for that to migrate current balance).

MarginPoolV2 is MarginPool, adding a mapping to keep track of deposited collateral that can be borrowed + this BorrowablePool implementation, what u think about that? @chuddster

haythem96:
> > > Left some comments, main concerns re figuring out which pool to redeem oToken from. I actually think upgrading the MarginPool and having one MarginPool that keep track of collateral that can be borrowed may be better
> > 
> > 
> > Also probably in that case whitelisting oToken buyer is not needed...
> 
> Something like 1 pool contract (but we need to migrate current pool funds to the new one, but can be done by adding a function in Controller that can only be called by owner for that to migrate current balance).
> 
> MarginPoolV2 is MarginPool, adding a mapping to keep track of deposited collateral that can be borrowed + this BorrowablePool implementation, what u think about that? @chuddster

Or if we want to keep the 2 different pools, I guess it make sense to update the oTokenImplementation and add an isBorrowable field

haythem96:
Closed by mistake...

haythem96:
> > > Left some comments, main concerns re figuring out which pool to redeem oToken from. I actually think upgrading the MarginPool and having one MarginPool that keep track of collateral that can be borrowed may be better
> > 
> > 
> > Also probably in that case whitelisting oToken buyer is not needed...
> 
> Something like 1 pool contract (but we need to migrate current pool funds to the new one, but can be done by adding a function in Controller that can only be called by owner for that to migrate current balance).
> 
> MarginPoolV2 is MarginPool, adding a mapping to keep track of deposited collateral that can be borrowed + this BorrowablePool implementation, what u think about that? @chuddster

If MM with 1oToken can't borrow collateral greater than the amount of collateral deposited to mint that 1oToken, right? I think it is safe to use the same MarginPool, and allow all oToken holder to redeem from the same pool, as we are locking the oToken from borrower, right?

contracts/core/BorrowableMarginPool.sol
chuddster: In principle I agree, but this would require a redeployment of Whitelist.sol with all of its existing whitelist state right? Think it makes sense to keep it in borrowable pool entirely to minimize the blast radius of this when deploying to prod.

contracts/core/BorrowableMarginPool.sol
chuddster: oTokens are 8 decimals. 

If the collateral token X for the oToken has decimals >= 8, we need to scale the oToken balance up to match "the equivalent borrowable in the asset" since each oToken is one underlying. Conversely if collateral token has decimals <8, we need to scale the oToken balance down to match how much in underlying is actually borrowable

chuddster:
> > > > Left some comments, main concerns re figuring out which pool to redeem oToken from. I actually think upgrading the MarginPool and having one MarginPool that keep track of collateral that can be borrowed may be better
> > > 
> > > 
> > > Also probably in that case whitelisting oToken buyer is not needed...
> > 
> > 
> > Something like 1 pool contract (but we need to migrate current pool funds to the new one, but can be done by adding a function in Controller that can only be called by owner for that to migrate current balance).
> > MarginPoolV2 is MarginPool, adding a mapping to keep track of deposited collateral that can be borrowed + this BorrowablePool implementation, what u think about that? @chuddster
> 
> If MM with 1oToken can't borrow collateral greater than the amount of collateral deposited to mint that 1oToken, right? I think it is safe to use the same MarginPool, and allow all oToken holder to redeem from the same pool(no need to whitelist oToken buyer or check which pool to redeem from), as we are locking the oToken from borrower, right?

@haythem96 I think some more context will be helpful for you here: we are trying to build this pool out for select ribbon treasury partners to increase the yield on the options since the collateral can be borrowed. We want to decouple this into an entirely new margin pool for three reasons:
1. this allows opyn to continue to be the owner of the MarginPool
2. this allows Ribbon to be owner of new MarginPool instance and decide who is whitelisted, the borrowing percentages, etcetera etcetera since the collateral comes only from ribbon vaults that we whitelist
3. upgrading original MarginPool with this code allows the owner to potentially whitelist borrowing of oTokens with ETH / BTC / USDC as collateral. These are oTokens minted by Ribbon retail vaults and StakeDao retail vaults that don't consent to their collateral being borrowed. This upgrade allows for that to be a possibility. Confining these access controls to an entirely separate MarginPool that only has select altcoins in it minimizes risk.

chuddster:
> Left some comments, main concerns re figuring out which pool to redeem oToken from. I actually think upgrading the MarginPool and having one MarginPool that keep track of collateral that can be borrowed may be better

contracts/core/Controller.sol
chuddster: The `_args.owner` that are whitelisted are exclusively the Ribbon DOV vault addresses themselves which generate the vault. We don't plan on having them be naked margin but if we do we will only be constraining ourselves which is fine. As for spread vault, referring to my comments [here](https://github.com/opynfinance/GammaProtocol/pull/459#discussion_r929257240) shouldn't MarginCalculator treat vault type 2 same as vault type 1 (spread/max loss vault).

contracts/core/BorrowableMarginPool.sol
chuddster: This was a specific decision since you are not exactly borrowing from the options vault, you are borrowing from the margin pool that the options vault deposits into. So "whitelistedBorrowableVault" when describing a DOV feels like a misnomer at worst and confusing at best.

contracts/core/Controller.sol
chuddster: Responded redeem logic concerns at bottom.

> Also is there a scenario where it is more profitable for the borrower to not repay?

It is always more profitable to not repay. That is why we make off-chain loan agreements with the market makers (taken care of by ribbon) - which is also a reason why we want to separate into another vault for this.

contracts/core/Controller.sol
haythem96: Yes I agree, but in more generic way for Gamma, it is better to keep it more flexible for other use cases, right? as this assume `_args.owner` will always only be a ribbon DOV

haythem96: I'm not sure I get `Also, making it _args.vaultType == 2 requires us to change the code for and redeploy all the vaults that plan to be using this.` ?

A solution like the below allow `owner` to open a spread vault, naked margin and borrowable one(assuming owner is whitelisted)

```
if (_args.vaultType == 2) {
  require(borrowablePool.isWhitelistedOptionsVault(_args.owner));
}
vaultType[_args.owner][vaultId] = _args.vaultType;
```

haythem96:
@chuddster  Most of it looks fine to me.

Based on the current implementation, agree on:
- Keeping the new whitelist functions in the new BorrowablePool, no need to upgrade Whitelist module...etc
- Current naming LGTM

I still think the vault type set when opening a vault should change, the proposed implementation in the comment is more generic and does not assume that user who will use the borrowable pool will not need to open spread/naked margin https://github.com/opynfinance/GammaProtocol/pull/459#discussion_r930842294

contracts/core/Controller.sol
haythem96: Do u meant if implementing my suggestion, you will need to redeploy RIbbon vault? 
If so, how where you planning to handle that with the initial implementation https://github.com/opynfinance/GammaProtocol/pull/459#discussion_r924390199 ?

contracts/core/BorrowableMarginPool.sol
```diff
@@ -0,0 +1,314 @@
...
+
+        uint256 collateralAssetAmount = _oTokenToCollateralAssetAmount(collateralAsset, _oTokenAmount);
+
+        borrowed[msg.sender][collateralAsset] = outstandingAssetBorrow.add(collateralAssetAmount);
+
+        // transfer _oTokenAmount of oToken from borrower to _pool
+        ERC20Interface(_oToken).safeTransferFrom(msg.sender, address(this), _oTokenAmount);
+        // transfer collateralAssetAmount of collateralAsset from pool to borrower
+        ERC20Interface(collateralAsset).safeTransfer(msg.sender, collateralAssetAmount);
+        emit Borrow(_oToken, collateralAsset, collateralAssetAmount, msg.sender);
```
haythem96, line 195: Maybe use the `transferToUser()` and `transferToPool()` to not mess up the pool accounting? wyt?

contracts/core/BorrowableMarginPool.sol
```diff
@@ -0,0 +1,314 @@
...
+
+        borrowed[_borrower][collateralAsset] = borrowed[_borrower][collateralAsset].sub(_collateralAmount);
+
+        uint256 oTokensToRedeem = _collateralAssetToOTokenAmount(collateralAsset, _collateralAmount);
+
+        // transfer _amount of collateralAsset from borrower to pool
+        ERC20Interface(collateralAsset).safeTransferFrom(_repayer, address(this), _collateralAmount);
+        // transfer oTokensToRedeem of oToken from pool to _borrower
+        ERC20Interface(_oToken).safeTransfer(_borrower, oTokensToRedeem);
+        emit Repay(_oToken, collateralAsset, _collateralAmount, _borrower, _repayer);
```
haythem96, line 282: Same here?

haythem96 (approved):
Overall LGTM, left few more comments. Would love also @aleone review

haythem96:
@akwritescode Can you please check circleci, because some jobs can't be executed from outside opyn org members.

haythem96:
When CI successfully run, based on coverage may need to add some tests, especially one to cover cases where Controller is involved

contracts/core/Controller.sol
chuddster: Yes in general I agree your approach is better design (letting owner pass it through) instead of overriding whatever value is passed. Currently we pass in vault type 0 in these vaults which will be overridden with this upgraded controller. This implementation does not require any vaults to be redeployed - we can just call setWhitelistedOptionsVault(true) on all existing addresses and it will override the vault type to 2. We already have 4 treasury vaults that are deployed and have funds in them and we would have to redeploy all of them and pass in vault type 2. This would require all of the treasuries to migrate which is undesirable. I agree its a bit hacky.

contracts/core/BorrowableMarginPool.sol
chuddster: yeah will add `assetBalance[_asset] = assetBalance[_asset].add(_amount);` and `assetBalance[_asset] = assetBalance[_asset].sub(_amount);`

chuddster: Not going to update asset balance for oToken because it wasn't really intended for that purpose wyt

chuddster:
> When CI successfully run, based on coverage may need to add some tests, especially one to cover cases where Controller is involved

can you elaborate?

chuddster:
> @chuddster Most of it looks fine to me.
> 
> Based on the current implementation, agree on:
> 
> * Keeping the new whitelist functions in the new BorrowablePool, no need to upgrade Whitelist module...etc
> * Current naming LGTM
> 
> I still think the vault type set when opening a vault should change, the proposed implementation in the comment is more generic and does not assume that user who will use the borrowable pool will not need to open spread/naked margin [#459 (comment)](https://github.com/opynfinance/GammaProtocol/pull/459#discussion_r930842294)

Yes in general I agree your approach is better design (letting owner pass it through) instead of overriding whatever value is passed. Currently we pass in vault type 0 in these vaults which will be overridden with this upgraded controller. This implementation does not require any vaults to be redeployed - we can just call setWhitelistedOptionsVault(true) on all existing addresses and it will override the vault type to 2. We already have 4 treasury vaults that are deployed and have funds in them and we would have to redeploy all of them and pass in vault type 2. This would require all of the treasuries to migrate to a new contract which is undesirable. I agree its a bit hacky.

The only vaults that will be able to open a margin vault in the borrowable pool are ribbon vaults (we will whitelist them) - so we are only constraining ourselves if we ever want to do a naked margin vault.

haythem96:
> > When CI successfully run, based on coverage may need to add some tests, especially one to cover cases where Controller is involved
> 
> can you elaborate?

Just meant when CI run, to check the coverage if there are some important missing paths not tested.

haythem96 (approved):

contracts/core/BorrowableMarginPool.sol
haythem96: wdym? I think it best to make sure the pool accounting always match the pool balance, even for oToken

DemolaJames (dismissed):

test/unit-tests/borrowableMarginPool.test.ts
```diff
@@ -0,0 +1,666 @@
...
+        'Borrowable amount mismatch',
+      )
+
+      await expectRevert(
+        marginPool.borrow(otoken.address, oTokenAmount.div(2).plus(1), { from: user1 }),
+        'MarginPool: Borrowing more than allocated',
+      )
+    })
+
+    it('should revert borrowing after borrowing full amount', async () => {
```
aleone, line 410: I think this is testing same as above, not after any borrow of full amount - its just a user with no balance.

test/unit-tests/marginCalculator.test.ts
```diff
@@ -252,6 +254,7 @@ contract('MarginCalculator', () => {
       })
       it('Should revert when vault contain more than 1 collateral asset', async () => {
         const vault = {
+          marginPool: ZERO_ADDR,
```
aleone, line 257: why did we add this?

test/unit-tests/marginCalculator.test.ts
```diff
@@ -238,6 +239,7 @@ contract('MarginCalculator', () => {
       })
       it('Should revert when vault contain more than 1 long asset', async () => {
         const vault = {
+          marginPool: ZERO_ADDR,
```
aleone, line 242: why did we add this?

test/unit-tests/marginCalculator.test.ts
```diff
@@ -224,6 +224,7 @@ contract('MarginCalculator', () => {
     describe('Vault prerequisites checks.', () => {
       it('Should revert when vault contain more than 1 short', async () => {
         const vault = {
+          marginPool: ZERO_ADDR,
```
aleone, line 227: why did we add this?

test/unit-tests/borrowableMarginPool.test.ts
```diff
@@ -0,0 +1,666 @@
...
+        new BigNumber(oTokenBalanceAfter).toString(),
+        'oToken amount mismatch',
+      )
+
+      assert.equal(
+        oTokenToUnderlying.toString(),
+        new BigNumber(amtBorrowableBefore).minus(amtBorrowableAfter).toString(),
+        'Borrowable amount mismatch',
+      )
+
```
aleone, line 465: the amtBorrowableAfter should also be 0

test/unit-tests/controller.test.ts
```diff
@@ -314,10 +320,17 @@ contract(
 
         const vaultCounterAfter = new BigNumber(await controllerProxy.getAccountVaultCounter(accountOwner1))
         assert.equal(vaultCounterAfter.minus(vaultCounterBefore).toString(), '1', 'vault counter after mismatch')
+        assert.equal(finalMarginPool, marginPool.address, 'vault margin pool mismatch')
       })
 
       it('should open vault from account operator', async () => {
+        await borrowableMarginPool.setOptionsVaultWhitelistedStatus(accountOwner1, true, { from: owner })
```
aleone, line 327: I don't like how in this test pretty much all of the accountOwner1 tests are now only done on the borrowable margin vault.  I'd prefer that we do on two different owners one that is borrowable and one that is regular and do tests on each.  Could even just create a 2nd unit test which repeats all the code but just whitelists the one account in one case and doesn't in 2nd.

test/unit-tests/controller.test.ts
```diff
@@ -2707,11 +2839,19 @@ contract(
             data: ZERO_ADDR,
           },
         ]
-        await usdc.approve(marginPool.address, collateralToDeposit, { from: accountOwner1 })
+        await borrowableMarginPool.setOTokenBuyerWhitelistedStatus(holder1, true, { from: owner })
+        finalMarginPool =
```
aleone, line 2843: We should test this flow, but it should be in its own file vs removing test functionality that exists (add vs change tests).

aleone (commented):
Generally looks good to me with a few comments throughout that would like a look at.  

Main comment is on the accounting and slight feeling of unease around how exercise occurs on the different vaults (but can't figure out an issue with it yet).  There are some tests where new borrowable margin pool is tested but testing of non-borrowable is removed, I'd like the test to be additive and not remove existing testing.  If we make changes to something in future, want to have both of them fully tested to prevent missing coverage/test cases etc.


I also think it makes sense to at least add 1 integration test flow w non-mocked contracts vs all in a unit test w mocks.

test/unit-tests/controller.test.ts
chuddster: I added a `controllerBorrowableMarginPoo.test.ts` in addition to the`controller.test.ts`. 

In `controllerBorrowableMarginPoo.test.ts` I test the changed flow.
In `controller.test.ts` I test the flow from this comment as well as from comment above (non borrowable accountOwner1)

test/unit-tests/borrowableMarginPool.test.ts
chuddster: Yes it is. Will delete.

test/unit-tests/marginCalculator.test.ts
chuddster: deleted

test/unit-tests/marginCalculator.test.ts
chuddster: deleted

test/unit-tests/borrowableMarginPool.test.ts
chuddster: set to 0.

test/unit-tests/controller.test.ts
```diff
@@ -314,10 +320,17 @@ contract(
 
         const vaultCounterAfter = new BigNumber(await controllerProxy.getAccountVaultCounter(accountOwner1))
         assert.equal(vaultCounterAfter.minus(vaultCounterBefore).toString(), '1', 'vault counter after mismatch')
+        assert.equal(finalMarginPool, marginPool.address, 'vault margin pool mismatch')
       })
 
       it('should open vault from account operator', async () => {
```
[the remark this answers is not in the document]
chuddster: removing as it does not belong in this test

contracts/core/BorrowableMarginPool.sol
```diff
@@ -0,0 +1,354 @@
...
+        borrowPCT[_collateral] = _borrowPCT;
+        emit SetBorrowPCT(_collateral, _borrowPCT);
+    }
+
+    /**
+     * @notice Lends out asset to market maker
+     * @param _oToken address of the oToken laying claim to the collateral assets
+     * @param _oTokenAmount amount of the oToken to post as collateral in exchange
+     */
+    function borrow(address _oToken, uint256 _oTokenAmount) public onlyWhitelistedBorrower {
```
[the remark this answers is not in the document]
chuddster: > However, someone can just send the oTokens that they have held to some other whitelisted address after borrowing, allowing the other party to borrow more. 

Yes, the following is possible: 

```
Whitelisted Borrower A has 100 oTokens, borrowPCT is 50%
Whitelisted Borrower B has 100 oTokens, borrowPCT is 100%

Borrower A borrowers 50 of underlying, contract locks 50 oTokens
Borrower A transfers rest of 50 oTokens to Borrower B
Borrower B has 150 oTokens, borrowPCT is 100%
Borrower B borrows 50 of underlying, contract locks 50 oTokens
Borrower B transfers 50 of underlying back to Borrower A
```

End result: Borrower A was able to circumvent the borrow limit by colluding with Borrower B who has not sacrificed anything in the process. We do rely on KYC/AML here and the reputation of these market makers to not do this. After all Borrower B is risking his reputation / not being able to borrow in the future by doing this. This loophole is easy to monitor and I don't see a way of getting around this on a smart contract level. 

> A MM can also sell their tokens allowing them to borrow more than they should be permitted to do, as they didn't have to put in collateralBorrowed/borrowPCT oTokens

What exactly do you mean? Selling oTokens prevents them from borrowing underlying.

> Is the depositing of oTokens just meant as a sort of receipt

Locking solves the two issues mentioned in anton's [comment](https://github.com/opynfinance/GammaProtocol/pull/459#discussion_r927660297). This does not solve the issue above but at least they cannot borrow more of the underlying than the oTokens they actually bought (essentially stealing from other whitelisted borrowers).

> Why not require them to put up 200 oTokens to borrow 100 underlyingAssets?

We could, but I don't see how it prevents getting around borrow limits in situation described above. Just can transfer the oTokens beforehand to some whitelisted market maker with a higher borrow limit.

test/unit-tests/marginCalculator.test.ts
chuddster: deleted

haythem96 (dismissed):

haythem96 (approved):

aleone (approved):
lgtm

